PR #21, opened 2025-04-25
update ontologies/science HTTP

### LINKED ISSUE #18: Eliminate use of pre-shared key
Depends on https://github.com/canvas-medical/canvas-plugins/pull/529

beaugunderson:
fixes #18

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,1 @@
+.coverage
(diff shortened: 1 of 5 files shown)


andhines (approved):
Thanks BG, looks like this preserves the conditional use of prod services, which is helpful for dev against local canvas.

PR #35, opened 2025-06-05
Add thread cleanup to each thread

beaugunderson:
this change runs thread cleanup at the conclusion of each thread, which prevents the database pool from being exhausted.

this needs a change deployed in `canvas-plugins` before it will work!

### DIFF
--- a/hyperscribe/handlers/commander.py
+++ b/hyperscribe/handlers/commander.py
@@ -7,6 +7,7 @@ from time import time
 from typing import Iterable
 
 import requests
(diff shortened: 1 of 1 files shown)


dbajet (approved):
Besides the comment related to the code style of the project, adding  a test for the new function, and changing the tests of the code where the new function is used should be done (I'll do it after the merge if needed).

dbajet:
> feel free to make those changes in this branch @dbajet

Will do when `canvas_sdk/utils/db.py` is part of the release, thanks!

PR #69, opened 2025-07-23
Updating enums and synthetic case upsert 

aaryanxshah:
I have replaced the enums + making sure the tests are consistent for these classes. Additionally, I have replicated the upsert and test_upsert logic for synthetic_case. 

Note: Most of the 59 file changes are from automatic linting. To actually see which files must be filtered/reviewed, please check files in these folders: `evaluations/datastores/postgres/`, `evaluations/structures/enums/`, and `evaluations/structures/records/` (along with the test files for these folders). 

To be precise, the files are:
1. `evaluations/datastores/postgres/synthetic_case.py` and `tests/evaluations/datastores/postgres/test_synthetic_case.py`
2. `evaluations/structures/enums/synthetic_case_x.py` and `tests/evaluations/structures/enums/test_synthetic_case_x.py`
3. `evaluations/structures/records/synthetic_case.py` and `tests/evaluations/structures/records/test_synthetic_case.py`

Also, if there is a more preferable way to remove the linted files for this PR (to avoid the massive list of changed files), happy to delete this and provide it based on the preferred method. Thank you for your review!

### DIFF
--- a/tests/evaluations/case_builders/test_synthetic_transcript_generator.py
+++ b/tests/evaluations/case_builders/test_synthetic_transcript_generator.py
@@ -54,14 +54,14 @@ def test__make_specifications(mock_choice, mock_sample, mock_randint, mock_unifo
         + ["anxious and talkative"]
     )
 
-    mock_sample.return_value = ["patient is frustrated", "patient is tearful"]
+    mock_sample.side_effect = [["patient is frustrated", "patient is tearful"]]
 
     specifications = tested._make_specifications()
     assert specifications["bucket"] == SyntheticCaseTurnBuckets.SHORT
     assert specifications["turn_total"] == 3
     assert specifications["speaker_sequence"] == ["Clinician", "Patient", "Patient"]
     assert specifications["ratio"] == 1.25
-    assert specifications["mood"] == mock_sample.return_value
+    assert specifications["mood"] == ["patient is frustrated", "patient is tearful"]
 
     assert mock_randint.mock_calls == [call(2, 4)]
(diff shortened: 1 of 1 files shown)


dbajet (changes requested):
Like said in Slack, this is bizarre that to see after the rebase the changes I did and committed as changes from the main branch? I guess some branch path convolution?

On the code and test, nothing to say :+1:

Nevertheless there are a few changes to consider.

1. One thing is missing is a changed `evaluations/datastores/postgres/hyperscribe.sql` to reflect the enums' changes
2. Also, now that the enums have been created, the corresponding constants should be replaced:
* `Constants.PATIENT_PERSONAS` -> `SyntheticCasePatientStyle`
* `Constants.MOOD_POOL` -> `SyntheticCaseMood`
...
3. The `SyntheticCase.speaker_sequence` is defined as string, while, based on the tests, it should be a list of strings. So, it should be changed as well as the underlying SQL (`text` -> `JSON`).

tests/evaluations/case_builders/test_synthetic_transcript_generator.py
```diff
@@ -58,7 +57,7 @@ def test__make_specifications(mock_choice, mock_sample, mock_randint, mock_unifo
     mock_sample.return_value = ["patient is frustrated", "patient is tearful"]
```
dbajet, line 57: no reason to not use `side_effect`:
```python
mock_sample.side_effect = [["patient is frustrated", "patient is tearful"]]
```

dbajet (approved):
All good :loudspeaker:
Thanks!

PR #85, opened 2025-08-25
rubric generator modifications

aaryanxshah:
Rubric generation (and downstream note grading) is adjusted to remove the `sense` variable, while maintaining a `Reward for` prefix for standardization.


tests/evaluations/case_builders/test_rubric_generator.py
```diff
@@ -94,7 +89,7 @@ def reset_mocks():
 
     result_system_lines, result_user_lines = tested.build_prompts(transcript, chart, canvas_context)
     expected_system_md5 = "a32a64e63443ef0f076080b5be3873d9"
-    expected_user_md5 = "f08e7813044d358e28018de214b8a512"
+    expected_user_md5 = "8640c4f7a0f9369d0f0a7af503b5dce5"
```
andhines, line 92: At first I thought this was for a user ID, and wasn't sure why it should change. Then I looked at the code to understand, and saw it is actually md5 of the user prompt content. Changing the variable name from `expected_user_md5` to `expected_user_prompt_md5` would avoid that confusion.

evaluations/case_builders/rubric_generator.py
```diff
@@ -77,10 +72,10 @@ def build_prompts(self, transcript: dict, chart: dict, canvas_context: dict) ->
             " 2. Decide what an ideal scribe must capture.",
             " 3. Produce the rubric as a JSON array of objects.",
             "Each object keys:",
-            ' - criterion (string) — must start with "Reward for" or "Penalize for"',
+            ' - criterion (string) — must start with "Reward for"',
```
andhines, line 75: Originally, the "Reward for" and "Penalize for" prefixes were added just to distinguish the sense of the criterion. But now since we're in just one direction, It may be best to have no such prefix at all, leaving it implied. Not crucial by any means. But something that comes to mind. Counter argument is that it's better to be explicit.

andhines (approved):
Comments are suggestions to consider for future changes.

PR #109, opened 2025-09-18
Adds trial staff option and improves launcher performance

andhines:


andhines:
high number of files changed is due to settings update

PR #139, opened 2025-10-15
updates prescription quantity format for discrete medications

alexiadowns-canvas:
**Problem**: Invalid prescription quantities for discrete medications (tablets, capsules) were formatted with unnecessary decimal places (e.g., "60.00" instead of "60"). 

**Root Cause**:  `.quantize(Decimal("0.01"))` was applied to all medication quantities, regardless of whether the medication was a discrete unit (tablets) or continuous measurement (liquids).

**Solution**: new classification system that determines if a medication is discrete or continuous based on the ncpdp quantity qualifier description, then uses the appropriate format.

### DIFF
--- a/hyperscribe/CANVAS_MANIFEST.json
+++ b/hyperscribe/CANVAS_MANIFEST.json
@@ -1,6 +1,6 @@
 {
   "sdk_version": "0.64.0",
-  "plugin_version": "2025-10-13 v0.2.128 (next a98779b)",
+  "plugin_version": "2025-10-14 v0.1.127 (main abc1234)",
   "name": "hyperscribe",
   "description": "Create commands based on the content of an audio discussion between a patient and a provider.",
   "url_permissions": [],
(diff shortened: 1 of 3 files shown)


tests/hyperscribe/commands/test_base_prescription.py
```diff
@@ -442,6 +465,34 @@ def reset_mocks():
...
+        type_to_dispense={"representative_ndc": "ndc1", "ncpdp_quantity_qualifier_code": "qualifier1"},
+        quantity_to_dispense=Decimal("60"),  # No .00 - integer format
+        refills=0,
+        note_to_pharmacist="Dispense 60 tablets",
+        sig="Take 2 tablets by mouth once daily",
+    )
+    assert command == expected
+    assert command.quantity_to_dispense == Decimal("60")  # Verify it's "60" not "60.00"
+    reset_mocks()
+
```
dbajet: As soon as a mock is used, a check of its calls should be performed, something like:
```python
    assert command.quantity_to_dispense == Decimal("60")  # Verify it's "60" not "60.00"

    calls = [...]
    assert demographic.mock_calls == calls
    calls = [...]
    assert chatter.mock_calls == calls
    
    reset_mocks()

```

alexiadowns-canvas: Thank you, that makes sense - added this for the changes I made!

andhines (approved):

PR #141, opened 2025-10-16
allow empty string for approximate date

alexiadowns-canvas:
**Problem**
Hyperscribe was incorrectly assigning the date of dictation to past surgical histories when no date was explicitly mentioned in the transcript.

**Root Cause** 
The issue occurred in the create_sdk_command_parameters() method in audio_interpreter.py where:
The system prompt included the current timestamp: "Please, note that now is 2025-10-06T09:20:40.116111"
The command_parameters() for SurgeryHistory specified: "approximateDate": "YYYY-MM-DD" without guidance on what to do when no date is available
The LLM, seeing no date in the transcript but having the current date in context, incorrectly defaulted to using today's date

This violated the instruction at line 341 of audio_interpreter.py: "you must restrict your response to information explicitly present in the transcript or prior instructions"

**Updates**
Changed `command_parameters()` methods in command classes with approximate date to explicitly instruct the LLM to leave dates empty when not mentioned: "YYYY-MM-DD or empty string if no date information is available in the transcript"

### DIFF
--- a/hyperscribe/commands/medical_history.py
+++ b/hyperscribe/commands/medical_history.py
@@ -73,10 +73,10 @@ class MedicalHistory(Base):
 
     def command_parameters(self) -> dict:
         return {
-            "keywords": "comma separated keywords of up to 5 synonyms of the condition",
-            "approximateStartDate": "YYYY-MM-DD or empty string if no date information is available in the transcript",
-            "approximateEndDate": "YYYY-MM-DD or empty string if no date information is available in the transcript",
-            "comments": "provided description of the patient specific history with the condition, as free text",
+            "keywords": "",
+            "approximateStartDate": None,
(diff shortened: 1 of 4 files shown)


dbajet:
This type of issue has been addressed in the command [hyperscribe/commands/immunization_statement.py](hyperscribe/commands/immunization_statement.py) by allowing the date to be `None` in the JSON Schema (see methods `command_parameters` and `command_parameters_schemas` [here](https://github.com/canvas-medical/canvas-hyperscribe/blob/53b271498dc11ecaeeee8061948eb998e11e7d53/hyperscribe/commands/immunization_statement.py#L95)), something to keep in mind when refactoring the `MedicalHistory` and `SurgeryHistory` commands.

PR #154, opened 2025-10-28
prevents selection of unlabeled questionnaire responses

alexiadowns-canvas:
**Feedback 4e64703a**
Hyperscribe selected the incorrect response in base_questionnaire.

**Root Cause**
The responses in the questionnaire were unlabeled, so an arbitrary response was selected.

**Expected Behavior**
On occasion, Hyperscribe will receive data from questionnaires without labeled responses. 
In this event, it should fetch complete questionnaire definitions from the Canvas SDK.
If for any reason responses are still unlabeled, they should be ignored (not selected). 

**Solutions**
Enhanced Option Label Handling ([base_questionnaire.py]
Added _fetch_complete_option_labels() method to query Canvas SDK for complete questionnaire definitions
Modified staged_command_extract() to use fallback logic: staged_label or full_label
Only filters out options if both sources (staged data AND full definition) have no label
Added error handling and logging for questionnaire fetching

### DIFF
--- a/hyperscribe/CANVAS_MANIFEST.json
+++ b/hyperscribe/CANVAS_MANIFEST.json
@@ -1,6 +1,6 @@
 {
   "sdk_version": "0.64.0",
-  "plugin_version": "2025-10-28 v0.2.133 (ad-4e64703a-2)",
+  "plugin_version": "2025-10-28 v0.2.134 (main)",
   "name": "hyperscribe",
   "description": "Create commands based on the content of an audio discussion between a patient and a provider.",
   "url_permissions": [],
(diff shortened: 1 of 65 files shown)


dbajet:
@alexiadowns-canvas , I don't know if you are waiting for another review or you are still working on it.

For what I see now.
(1) The `_fetch_complete_option_labels` method is making queries to the database.
Actually, we want to gather all database queries in the `LimitedCache` class, this way we can run the code in an isolated context.

(2) The `_fetch_complete_option_labels` is just retrieving the same data that is inthe `data` parameter of the `staged_command_extract` method.
Or at least it should be the same data.
If it is not, then there is a bigger problem (outside of the Hyperscribe plugin) that should be investigated and reported.

(3) Most importantly, I am not comfortable with this approach of ignoring an option because the label is empty (whatever is the method used then) without better context.
Like in (2), if options without label are a real issue, we need to investigate the root cause of these empty labels.
If they are empty on purpose, maybe ignoring the options is not the correct approach.

Any comment on this @andhines?

PR #162, opened 2025-11-04
Inclusion and exclusion criteria for HPI

alexiadowns-canvas:
**Feedback**
Customers have requested more information in the HPI, related to and describing patient symptoms (onset, duration, character, timing and severity) in a narrative format.

If there is not enough information in the transcript, the LLM adds an explanation:
_"The condition was mentioned during the clinical encounter, but no specific patient history or details were provided,"_

**Changes**
Added a shortened version of CMS's definition of HPI 
CMS: (location; quality; severity; duration; timing; context; modifying factors; associated signs and symptoms)

Added "and no instruction in the lack of" framework from other commands to let the LLM know that it doesn't have to add an explanation

Note: this change is made with instance-specific prompt amendments in mind, in order to clarify inclusion criteria for a more verbose response that can later be filtered


alexiadowns-canvas:
closed based on recent testing with custom prompts

PR #184, opened 2025-11-27
resumable experiment runner b/c of tunnel timeouts

aaryanxshah:
added a --resume flag to allow for in-progress experiment run to pick up where it leaves off rather than restarting. this queries the remaining jobs that have been set forth to do so and then start in that process.


scripts/experiment_runner.py
```diff
@@ -108,8 +150,18 @@ def run(cls) -> None:
                 worker.start()
                 note_grader_workers.append(worker)
 
-            for job in cls._generate_jobs(args.experiment_id, clone_repository):
+            jobs_added = 0
+
+            for job in cls._generate_jobs(
+                args.experiment_id, clone_repository, completed_jobs if args.resume else set()
```
dbajet: `completed_jobs` is different of `set()` only if `args.resume` is set to `True` (lines 120-123`, so here `if args.resume else set()` doesn't help.

scripts/experiment_runner.py
```diff
@@ -143,10 +197,32 @@ def _generate_jobs(cls, experiment_id: int, repository: Path) -> Generator[Exper
             print("Experiment has no models")
             return
 
+        if completed_jobs is None:
```
dbajet: This will disappear when changing `completed_jobs` as a required parameter.

dbajet (changes requested):
Implementing this feature is great!

But there are a few concerns:
- the SQL is incorrect, 
- using named tuple helps
- optional parameters should be avoided

dbajet (changes requested):
Thank for the changes!


There are still a few things:
- the SQL should be move to the dedicated class
- there is no test file for `evaluations/structures/completed_job.py` (it may be very similar to `tests/evaluations/structures/test_case_runner_job.py` and `tests/hyperscribe/structures/test_instruction.py`
- the tests are failing ([see here](https://github.com/canvas-medical/canvas-hyperscribe/actions/runs/19878270126/job/56970554545?pr=184))

Did you install the pre-commit?
If not, run this: `uv run pre-commit install --install-hooks`

PR #186, opened 2025-12-02
restore msgLower variable and ensure websocket reconnection

alexiadowns-canvas:
Restores the variable that was removed in commit 269b0d5
Allows addEventMessage() function to process events correctly


alexiadowns-canvas:
> It feels like we may need some UI tests to avoid losing some code 😑

Yes, I agree... 😞  I updated my pre-commit UAT process to prevent this and created this issue for new tests:
https://github.com/canvas-medical/canvas-hyperscribe/issues/187

Issue #200, opened 2026-02-02
Stop originating commands for denial of medical history


alexiadowns-canvas:
<img width="762" height="172" alt="Image" src="https://github.com/user-attachments/assets/970a29a9-2c09-4541-9606-790638b72abc" />

PR #201, opened 2026-02-06
handles denials for pmh and psh

alexiadowns-canvas:
Addressing #200 

- Two prompt changes to prevent Hyperscribe from originating Past Medical History and Past Surgical History commands when a patient denies a condition or surgery. 
- Updated instruction_description() in both medical_history.py and surgery_history.py to explicitly require patient confirmation and reject denied/negated history.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,5 @@ evaluations.db
 local_env.sh
 **/.DS_Store
 .DS_Store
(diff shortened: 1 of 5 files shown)


dbajet (approved):
It sounds good.
Is there any other command that would benefit of this?

Issue #189, opened 2025-12-10
Note templates


andhines:
Add a new plugin or new handler(s) (need to decide) that enables note templates that are more compatible with Hyperscribe. Notably, they must yield a permission data structure that Hyperscribe can comply with, specifying which commands, fields, subfields Hyperscribe is allowed to edit.

PR #204, opened 2026-02-08
Add note template permission integration                                                          

alexiadowns-canvas:
Addresses #189 by adding a note template permission integration                                                          
                                                                                                    
Adds support for note_templates cached permission system. 
When a note has a template applied, Hyperscribe now respects per-command and per-field edit permissions, {add:} instructions, and {lit:} framework content set by the template.                                   

**How it Works**
  - Permission checks: Before generating content for a command field, checks the shared cache for template permissions. Locked fields are skipped entirely.
  - Template frameworks: When a field has a plugin_edit_framework, Hyperscribe fills in the
  framework structure rather than generating free-form prose. Falls back to detecting structure in existing note content.
- {add:} instructions: Template-defined instructions (e.g. "symptoms", "duration") guide the LLM to enhance content with specific details.
- {lit:} markers: Literal text in frameworks is preserved exactly as written and stripped before sending to the LLM, then restored in the output.

**Changes**

- template_constants.py  Maps Hyperscribe class names to Canvas command types 
- template_permissions.py Reads permissions from the shared note_template_permissions cache
- template_mixin.py TemplateIntegrationMixin with resolve_field(), fill_template_content(), and
  framework resolution logic
- base.py — Inherits the mixin; minimal changes (3 lines added to the original file)
  - All 14 command files updated to use permission checks

**Plugin Interactions**
Both plugins use get_cache_client(driver="plugins", prefix="note_template_permissions") to share a cache namespace, bypassing the plugin-scoped get_cache().

### DIFF
--- a/hyperscribe/commands/base.py
+++ b/hyperscribe/commands/base.py
@@ -22,13 +22,13 @@ class Base:
         settings: Settings,
         cache: LimitedCache,
         identification: IdentificationParameters,
-        permissions: TemplatePermissions | None = None,
+        permissions: TemplatePermissions,
     ):
         self.settings = settings
         self.identification = identification
         self.cache = cache
         self._arguments_code2description: dict[str, str] = {}
-        self.permissions = permissions or TemplatePermissions(identification.note_uuid)
+        self.permissions = permissions
 
     @classmethod
     def class_name(cls) -> str:
--- a/tests/hyperscribe/commands/test_base.py
+++ b/tests/hyperscribe/commands/test_base.py
@@ -50,7 +50,7 @@ def helper_instance() -> Base:
         provider_uuid="providerUuid",
         canvas_instance="canvasInstance",
     )
-    return Base(settings, cache, identification)
+    return Base(settings, cache, identification, TemplatePermissions("noteUuid"))
 
 
 def test___init__():
@@ -78,13 +78,13 @@ def test___init__():
         provider_uuid="providerUuid",
         canvas_instance="canvasInstance",
     )
-    tested = Base(settings, cache, identification)
+    permissions = TemplatePermissions("noteUuid")
+    tested = Base(settings, cache, identification, permissions)
     assert tested.settings == settings
     assert tested.identification == identification
     assert tested.cache == cache
     assert tested._arguments_code2description == {}
-    assert isinstance(tested.permissions, TemplatePermissions)
-    assert tested.permissions.note_uuid == "noteUuid"
+    assert tested.permissions is permissions
 
 
 def test_class_name():
--- a/tests/hyperscribe/commands/test_update_goal.py
+++ b/tests/hyperscribe/commands/test_update_goal.py
@@ -13,6 +13,7 @@ from hyperscribe.structures.instruction_with_command import InstructionWithComma
 from hyperscribe.structures.instruction_with_parameters import InstructionWithParameters
 from hyperscribe.structures.settings import Settings
 from hyperscribe.structures.vendor_key import VendorKey
+from hyperscribe.libraries.template_permissions import TemplatePermissions
 
 
 def helper_instance() -> UpdateGoal:
@@ -40,7 +41,7 @@ def helper_instance() -> UpdateGoal:
         provider_uuid="providerUuid",
         canvas_instance="canvasInstance",
     )
-    return UpdateGoal(settings, cache, identification)
+    return UpdateGoal(settings, cache, identification, TemplatePermissions("noteUuid"))
 
 
 def test_class():
--- a/tests/hyperscribe/libraries/test_audio_interpreter.py
+++ b/tests/hyperscribe/libraries/test_audio_interpreter.py
@@ -1,5 +1,5 @@
 from datetime import datetime, timezone
-from unittest.mock import ANY, patch, MagicMock, call
+from unittest.mock import patch, MagicMock, call
 
 import pytest
 
@@ -29,8 +29,12 @@ from tests.helper import MockClass
 def helper_instance(mocks, with_audit) -> tuple[AudioInterpreter, Settings, AwsS3Credentials, LimitedCache]:
     def reset_mocks():
         command_list.reset_mocks()
+        template_permissions.reset_mock()
 
-    with patch.object(ImplementedCommands, "command_list") as command_list:
+    with (
+        patch.object(ImplementedCommands, "command_list") as command_list,
+        patch("hyperscribe.libraries.audio_interpreter.TemplatePermissions") as template_permissions,
+    ):
         settings = Settings(
             llm_text=VendorKey(vendor="textVendor", api_key="textKey"),
             llm_audio=VendorKey(vendor="audioVendor", api_key="audioKey"),
@@ -60,6 +64,7 @@ def helper_instance(mocks, with_audit) -> tuple[AudioInterpreter, Settings, AwsS
                 mocks[4].return_value.is_available.side_effect = [True]
 
         command_list.side_effect = [mocks]
+        template_permissions.side_effect = ["templatePermissionsInstance"]
 
         cache = LimitedCache("patientUuid", "providerUuid", {})
         cache._demographic = "thePatientDemographic"
@@ -72,17 +77,21 @@ def helper_instance(mocks, with_audit) -> tuple[AudioInterpreter, Settings, AwsS
         instance = AudioInterpreter(settings, aws_s3, cache, identification)
         calls = [call()]
         assert command_list.mock_calls == calls
+        calls = [call("noteUuid")]
+        assert template_permissions.mock_calls == calls
         reset_mocks()
 
         return instance, settings, aws_s3, cache
 
 
+@patch("hyperscribe.libraries.audio_interpreter.TemplatePermissions")
 @patch.object(ImplementedCommands, "command_list")
-def test___init__(command_list):
+def test___init__(command_list, template_permissions):
     mocks = [MagicMock(), MagicMock(), MagicMock(), MagicMock(), MagicMock()]
 
     def reset_mocks():
         command_list.reset_mocks()
+        template_permissions.reset_mock()
         for item in mocks:
             item.reset_mock()
 
@@ -118,6 +127,7 @@ def test___init__(command_list):
     mocks[3].return_value.class_name.side_effect = ["CommandD", "CommandD"]
     mocks[4].return_value.class_name.side_effect = ["CommandE", "CommandE"]
     command_list.side_effect = [mocks]
+    template_permissions.side_effect = ["templatePermissionsInstance"]
 
     cache = LimitedCache("patientUuid", "providerUuid", {})
     identification = IdentificationParameters(
@@ -133,11 +143,13 @@ def test___init__(command_list):
     assert instance.cache == cache
     calls = [call()]
     assert command_list.mock_calls == calls
+    calls = [call("noteUuid")]
+    assert template_permissions.mock_calls == calls
     for idx, mock in enumerate(mocks):
         # With walrus operator optimization, class_name() is called only once
         # (assigned via := in policy check) and reused for dict key
         calls = [
-            call(settings, cache, identification, ANY),
+            call(settings, cache, identification, "templatePermissionsInstance"),
             call().__bool__(),
             call().class_name(),
             call().is_available(),
@@ -148,8 +160,9 @@ def test___init__(command_list):
     reset_mocks()
 
 
+@patch("hyperscribe.libraries.audio_interpreter.TemplatePermissions")
 @patch.object(ImplementedCommands, "command_list")
-def test___init___with_template_filtering(command_list):
+def test___init___with_template_filtering(command_list, template_permissions):
     """Test that commands locked by template permissions are excluded from _command_context."""
     mocks = [MagicMock(), MagicMock(), MagicMock()]
 
@@ -184,6 +197,7 @@ def test___init___with_template_filtering(command_list):
     mocks[1].return_value.class_name.side_effect = ["CommandB", "CommandB"]
     mocks[2].return_value.class_name.side_effect = ["CommandC", "CommandC"]
     command_list.side_effect = [mocks]
+    template_permissions.side_effect = ["templatePermissionsInstance"]
 
     cache = LimitedCache("patientUuid", "providerUuid", {})
     identification = IdentificationParameters(
@@ -203,9 +217,11 @@ def test___init___with_template_filtering(command_list):
 
     calls = [call()]
     assert command_list.mock_calls == calls
+    calls = [call("noteUuid")]
+    assert template_permissions.mock_calls == calls
     for mock in mocks:
         calls = [
-            call(settings, cache, identification, ANY),
+            call(settings, cache, identification, "templatePermissionsInstance"),
             call().__bool__(),
             call().class_name(),
             call().is_available(),
@@ -273,7 +289,7 @@ def test_common_instructions():
         assert result == expected
         for idx, mock in enumerate(mocks):
             calls = [
-                call(settings, cache, tested.identification, ANY),
+                call(settings, cache, tested.identification, "templatePermissionsInstance"),
                 call().__bool__(),
                 call().class_name(),
                 call().is_available(),
@@ -324,7 +340,7 @@ def test_instruction_constraints():
         assert result == expected
         for idx, mock in enumerate(mocks):
             calls = [
-                call(settings, cache, tested.identification, ANY),
+                call(settings, cache, tested.identification, "templatePermissionsInstance"),
                 call().__bool__(),
                 call().class_name(),
                 call().is_available(),
@@ -377,7 +393,7 @@ def test_command_structures(questionnaire_command_name_list):
         assert questionnaire_command_name_list.mock_calls == calls
         for idx, mock in enumerate(mocks):
             calls = [
-                call(settings, cache, tested.identification, ANY),
+                call(settings, cache, tested.identification, "templatePermissionsInstance"),
                 call().__bool__(),
                 call().class_name(),
                 call().is_available(),
@@ -430,7 +446,7 @@ def test_command_schema(questionnaire_command_name_list):
         assert questionnaire_command_name_list.mock_calls == calls
         for idx, mock in enumerate(mocks):
             calls = [
-                call(settings, cache, tested.identification, ANY),
+                call(settings, cache, tested.identification, "templatePermissionsInstance"),
                 call().__bool__(),
                 call().class_name(),
                 call().is_available(),
@@ -1101,7 +1117,7 @@ def test_detect_sections(json_schema, chatter, memory_log):
     assert memory_log.mock_calls == calls
     for idx, mock in enumerate(mocks):
         calls = [
-            call(settings, cache, tested.identification, ANY),
+            call(settings, cache, tested.identification, "templatePermissionsInstance"),
             call().__bool__(),
             call().class_name(),
             call().is_available(),
@@ -1529,7 +1545,7 @@ def test_detect_instructions_flat(json_schema, instruction_constraints, chatter,
     assert memory_log.mock_calls == calls
     for idx, mock in enumerate(mocks):
         calls = [
-            call(settings, cache, tested.identification, ANY),
+            call(settings, cache, tested.identification, "templatePermissionsInstance"),
             call().__bool__(),
             call().class_name(),
             call().is_available(),
@@ -2046,7 +2062,7 @@ def test_create_sdk_command_from(chatter, memory_log, progress):
         assert progress.mock_calls == calls
         for idx, mock in enumerate(mocks):
             calls = [
-                call(settings, cache, tested.identification, ANY),
+                call(settings, cache, tested.identification, "templatePermissionsInstance"),
                 call().__bool__(),
                 call().class_name(),
                 call().is_available(),
@@ -2142,7 +2158,7 @@ def test_update_questionnaire(chatter, memory_log):
         assert memory_log.mock_calls == calls
         for idx, mock in enumerate(command_mocks):
             calls = [
-                call(settings, cache, tested.identification, ANY),
+                call(settings, cache, tested.identification, "templatePermissionsInstance"),
                 call().__bool__(),
                 call().class_name(),
                 call().is_available(),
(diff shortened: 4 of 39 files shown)


hyperscribe/libraries/template_constants.py
```diff
@@ -0,0 +1,53 @@
...
+    "ReasonForVisit": "reasonForVisit",
+    "Refer": "refer",
+    "Task": "task",
+    "UpdateDiagnose": "updateDiagnosis",
+    "UpdateGoal": "updateGoal",
+}
+
+
+def get_schema_key(hyperscribe_class_name: str) -> str | None:
+    """Convert a Hyperscribe class name to its Canvas SDK schema_key, or None if not mapped."""
```
dbajet: The [Base](hyperscribe/commands/base.py) class, thus all Hyperscribe commands, has a `schema_key` method providing the Canvas SDK schema_keys.
They are all gathered in the method `schema_key2instruction` of the [ImplementedCommands](hyperscribe/libraries/implemented_commands.py).

Adding to the class a method `template_commands` listing the allowed commands would make the function `get_schema_key` useless and keep related things together:
```python
    @classmethod
    def template_command_list(cls) -> list[Type[Base]]:
        return [
            Allergy,
            Assess,
            CloseGoal,
            Diagnose,
            FamilyHistory,
            FollowUp,
            Goal,
            HistoryOfPresentIllness,
            Instruct,
            LabOrder,
            MedicalHistory,
            Perform,
            Plan,
            ReasonForVisit,
            Refer,
            Task,
            UpdateDiagnose,
            UpdateGoal,
        ]

```

hyperscribe/libraries/template_constants.py
```diff
@@ -0,0 +1,53 @@
...
+    return HYPERSCRIBE_TO_SCHEMA_KEY.get(hyperscribe_class_name)
```
dbajet: Additional note: `get_schema_key` defined here is used once in [TemplateIntegrationMixin](hyperscribe/commands/template_mixin.py).

The sub-classes of [Base](hyperscribe/commands/base.py) implements `schema_key` laready, so `get_schema_key` is just adding complexity by repetition.

hyperscribe/commands/diagnose.py
```diff
@@ -32,21 +32,31 @@ def command_from_json(
...
         chatter: LlmBase,
     ) -> InstructionWithCommand | None:
+        # Get field values with template permission checks
+        background = self.resolve_field("background", instruction.parameters["rationale"], instruction, chatter)
+        today_assessment = self.resolve_field(
+            "today_assessment", instruction.parameters["assessment"], instruction, chatter
+        )
+
+        # If neither field can be edited, skip this command
+        if background is None and today_assessment is None:
```
dbajet: Same comment that in the [Assess command](https://github.com/canvas-medical/canvas-hyperscribe/pull/204/changes#r2783522763).

hyperscribe/commands/refer.py
```diff
@@ -58,6 +58,16 @@ def command_from_json(
...
     ) -> InstructionWithCommand | None:
+        # Get field values with template permission checks
+        notes_to_specialist = self.resolve_field(
+            "notes_to_specialist", instruction.parameters["notesToSpecialist"], instruction, chatter
+        )
+        comment = self.resolve_field("comment", instruction.parameters["comment"], instruction, chatter)
+
+        # If neither field can be edited, skip this command
+        if notes_to_specialist is None and comment is None:
+            return None
```
dbajet: Same comment that in the [Assess command](https://github.com/canvas-medical/canvas-hyperscribe/pull/204/changes#r2783522763).

hyperscribe/commands/task.py
```diff
@@ -125,26 +125,34 @@ def command_from_json(
         instruction: InstructionWithParameters,
         chatter: LlmBase,
     ) -> InstructionWithCommand | None:
+        # Get field values with template permission checks
+        title = self.resolve_field("title", instruction.parameters["title"], instruction, chatter)
+        comment = self.resolve_field("comment", instruction.parameters["comment"], instruction, chatter)
+
+        # If neither field can be edited, skip this command
+        if title is None and comment is None:
+            return None
```
dbajet: Same comment that in the [Assess command](https://github.com/canvas-medical/canvas-hyperscribe/pull/204/changes#r2783522763).

hyperscribe/libraries/template_permissions.py
```diff
@@ -0,0 +1,128 @@
...
+from hyperscribe.libraries.template_constants import (
+    COMMAND_PERMISSIONS_KEY_PREFIX,
+    get_command_type,
+)
+
+
+# Shared prefix must match what brigade_note_templates/template_cache.py writes with.
+# We bypass the plugin-scoped get_cache() (which auto-prefixes with the calling
+# plugin's name) so that both plugins read/write the same cache keys.
+SHARED_CACHE_PREFIX = "note_template_permissions"
```
dbajet: the constant could be added to the [libraries/constants.py](hyperscribe/libraries/constants.py) file.

hyperscribe/commands/template_mixin.py
```diff
@@ -0,0 +1,283 @@
...
+            json.dumps([{"enhancedContent": "the enhanced content here"}]),
+            "```",
+            "",
+        ]
+
+        chatter.reset_prompts()
+        if response := chatter.single_conversation(system_prompt, user_prompt, schemas, instruction):
+            enhanced = str(response[0].get("enhancedContent", content))
+            if enhanced:
+                return enhanced
```
dbajet: this could be replaced with
```python
            if enhanced := str(response[0].get("enhancedContent")):
                return enhanced
```

hyperscribe/commands/template_mixin.py
```diff
@@ -0,0 +1,283 @@
...
+            return False
+
+        lines = content.strip().split("\n")
+
+        # Look for section header patterns (lines ending with colon, or "Header: content" patterns)
+        header_pattern_count = 0
+        for line in lines:
+            line = line.strip()
+            # Pattern: "Something something:" at start of line (section header)
+            if line.endswith(":") and len(line) > 5:
```
dbajet: same `magic value` (see [comment](https://github.com/canvas-medical/canvas-hyperscribe/pull/204/changes#r2783686052))

hyperscribe/commands/template_mixin.py
```diff
@@ -0,0 +1,283 @@
...
+                header_pattern_count += 1
+            # Pattern: "Header: " followed by content on same line
+            elif ": " in line and line.index(": ") < 60:
```
dbajet: same `magic value` (see [comment](https://github.com/canvas-medical/canvas-hyperscribe/pull/204/changes#r2783686052))

hyperscribe/commands/template_mixin.py
```diff
@@ -0,0 +1,283 @@
...
+                prefix = line.split(": ")[0]
+                # Check if prefix looks like a header (starts with capital, reasonable length)
+                if prefix and prefix[0].isupper() and 3 <= len(prefix) <= 60:
```
dbajet: same `magic values` (see [comment](https://github.com/canvas-medical/canvas-hyperscribe/pull/204/changes#r2783686052))

hyperscribe/commands/base.py
```diff
@@ -12,12 +14,13 @@
...
 
 
-class Base:
+class Base(TemplateIntegrationMixin):
     def __init__(self, settings: Settings, cache: LimitedCache, identification: IdentificationParameters):
         self.settings = settings
         self.identification = identification
         self.cache = cache
         self._arguments_code2description: dict[str, str] = {}
+        self._template_permissions: TemplatePermissions | None = None
```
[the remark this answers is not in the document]
dbajet: The previous comment remains: 
- is there a reason to not instantiate the `_template_permissions` immediately? 
- can't `_template_permissions` be public? (i.e. `template_permissions`)

hyperscribe/commands/base.py
```diff
@@ -184,3 +194,188 @@ def instruction_constraints(self) -> str:
...
+    # -- Template integration ------------------------------------------------
+
+    @property
+    def template_permissions(self) -> TemplatePermissions:
+        """Lazily initialize and return the TemplatePermissions instance."""
+        if self._template_permissions is None:
+            self._template_permissions = TemplatePermissions(self.identification.note_uuid)
+        return self._template_permissions
+
+    def can_edit_field(self, field_name: str) -> bool:
```
dbajet: As discussed in the main comment, this method should just be the `can_edit_field` implemented in `TemplatePermissions`.

hyperscribe/libraries/template_permissions.py
dbajet: ### Design Issue

As said in the main comment of this review, the class `TemplatePermissions` holds few methods (`can_edit_command`, `can_edit_field`, `get_add_instructions` and `get_edit_framework`) that should be in the `Base` class. 
This will:
1. solve the circular import issue, and 
2. improve the design


### Caching Issue

The second important issue is that the same permissions are retrieved and hold for each command, regardless of the 'lazy' nature of the mechanism.
One way to solve this is to create a singleton of this class, hopefully now just managing the permissions.
Everything, except the singleton itself should be in the class to respect the OOP approach of the project.
The singleton is linked to the note 'uuid' and should live the time of the execution of the cycle only.


### Comments

Also, most of the comments of the previous review related to this file remain.

hyperscribe/libraries/implemented_commands.py
```diff
@@ -95,3 +95,27 @@ def command_list(cls) -> list[Type[Base]]:
     @classmethod
     def schema_key2instruction(cls) -> dict[str, str]:
         return {command_class.schema_key(): command_class.class_name() for command_class in cls.command_list()}
+
+    @classmethod
+    def template_command_list(cls) -> list[Type[Base]]:
```
dbajet: This method has been added following the previous review but is not used because the part that would have used it, no longer exists (`get_current_note_content` I think).

So the method could be safely removed.

dbajet (changes requested):
Thank you  @alexiadowns-canvas for the changes as they improve the code :pray:


### Missed comments

A bunch of comments have just been missed.

For example, all comments related to the fact that a lot of computation could be preventing by detecting early that some or all fields of a command cannot be edited  (e.g. for [hyperscribe/commands/assess.py](https://github.com/canvas-medical/canvas-hyperscribe/pull/204/changes#r2783522763), [hyperscribe/commands/close_goal.py](https://github.com/canvas-medical/canvas-hyperscribe/pull/204/changes#r2783531454)...).

These is perfectly fine, and expected when incorrect or invalid, to reject suggestions and requests, but there should be an explicit rational to it.


### Running the tests:

When running `uv run pytest tests/ --cov=.`, it doesn't end with 100%:
```shell
hyperscribe/libraries/template_permissions.py 98.21%   11
```

The culprit is:
```python
if TYPE_CHECKING:
    from hyperscribe.commands.base import Base
```

This `if TYPE_CHECKING:` is used here as a trick to break a circular import.
Indeed, more precisely, both modules are dependent of each other (`Base` imports `TemplatePermissions`, and `TemplatePermissions` imports `Base`).

###### why 100% is important

First, as a general comment, when the coverage is not 100%, one has to know very well the code to understand quickly if the missing coverage is critical or not.
Second, quickly from 2%, it will become 3, then 7, 11, 17... and tests won't matter anymore.

100% coverage is not a goal _per se_: it is helping to quickly identify what tests/scenario are missing, this is not a constraint for some purity of the tests.

###### dirty solution

The `if TYPE_CHECKING:` is, while a common practice, a dirty way (in my opinion) to circunvent a design issue.

The impacted methods (`can_edit_command`, `can_edit_field`, `get_add_instructions` and `get_edit_framework`) are already part of the `Base` class through shortcuts.
For example:
```python
def can_edit_field(self, field_name: str) -> bool:
  return self.template_permissions.can_edit_field(type(self), field_name)
```

Remove the `command_class: type[Base]` parameter and they can easily become instance methods.

That will also simplify the code, for example, in [AudioInterpreter](hyperscribe/libraries/audio_interpreter.py)
```python
        self._command_context = {
            class_name: instance
            for command_class in ImplementedCommands.command_list()
            if (instance := command_class(settings, cache, identification))
            and self.settings.commands_policy.is_allowed(class_name := instance.class_name())
            and instance.is_available()
            and self._template_permissions.can_edit_command(command_class)
        }
```
would become:
```python
        self._command_context = {
            class_name: instance
            for command_class in ImplementedCommands.command_list()
            if (instance := command_class(settings, cache, identification))
            and self.settings.commands_policy.is_allowed(class_name := instance.class_name())
            and instance.is_available()
            and instance.can_edit_command()
        }
```

The `TemplatePermissions` would only keep the `_permissions_cache` management.

##### singleton option

For a specific note and cycle, all the instances, i.e. for each command, of the `TemplatePermissions` class retrieve and hold the same permissions, regardless of the 'lazy' nature of the mechanism.

Creating a singleton, based on the note and cycle, of this class would simplify and be more efficient (see this [comment](https://github.com/canvas-medical/canvas-hyperscribe/pull/204/changes#r2793295795)).


### Mocks in tests

This is a general comments related to the tests.

Because of the expected changes, tests haven't been really reviewed.

But that said, and this is mentioned in [tests/hyperscribe/commands/test_assess.py](tests/hyperscribe/commands/test_assess.py), using a mock without checking its calls (through `mock_calls`) makes any test incomplete.

And, this increase the risk of incoherent behaviors or bugs.

Even if this example seems obvious, consider a mock called with `delete <file>` rather than `print <file>`, that won't be caught if the `mock_calls` is not checked.

This is valid for __all__ mocks.


### Tests format

This test is valid.
```python
def test_command_type():
    assert SurgeryHistory.command_type() == "PastSurgicalHistoryCommand"
```

But, following the project conventions, the test would be:
```python
def test_command_type():
    tested = SurgeryHistory
    result = tested.command_type()
    expected = "SurgeryHistory"
    assert result == expected
```

The benefit of conventions and similar patterns is that exceptions - valid, justified, intentional or not - are easier to pinpoint, making the reviews and maintenance easier as well.

hyperscribe/libraries/template_permissions.py
```diff
@@ -0,0 +1,128 @@
...
+
+try:
+    from canvas_sdk.caching.client import get_cache as _get_cache_client
+except ImportError:
+    _get_cache_client = None
```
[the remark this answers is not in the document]
alexiadowns-canvas: You're right - this import doesn't fail so this is unnecessary complexity and was added as a result of confusion between the plugin-scoped cache and the general cache. Removing!

hyperscribe/libraries/implemented_commands.py
alexiadowns-canvas: removed!

tests/hyperscribe/commands/test_base.py
```diff
@@ -449,3 +460,736 @@ def test_is_available():
...
+
+
+# =========================================================================
+# Template integration tests
+# =========================================================================
+
+FIXED_DATE = datetime(2025, 11, 4, 4, 55, 21, 12346, tzinfo=timezone.utc)
+
+
+def make_instruction(**overrides) -> InstructionWithParameters:
```
dbajet: Good idea!

tests/hyperscribe/commands/test_base.py
```diff
@@ -449,3 +460,736 @@ def test_is_available():
...
+    """Common setup for tests that exercise the LLM path."""
+    mock_json_schema.get.return_value = [{"type": "array"}]
+    mock_demographic.return_value = "theDemographic"
+    mock_datetime.now.return_value = FIXED_DATE
+
+
+# -- can_edit_command ------------------------------------------------------
+
+
+class TestCanEditCommand:
```
dbajet: Using test class... is not the convention of the project.
Not a big deal, but conventions are like patterns, they help.

tests/hyperscribe/commands/test_base.py
```diff
@@ -449,3 +460,736 @@ def test_is_available():
...
+    assert mock_json_schema.mock_calls == [call.get(["template_enhanced_content"])]
+    assert mock_demographic.mock_calls == [call(False)]
+    assert mock_datetime.mock_calls == [call.now()]
+    assert (
+        chatter.mock_calls
+        == [
+            call.reset_prompts(),
+            call.single_conversation(
+                chatter.reset_prompts.call_args,  # placeholder — checked via called
+                chatter.single_conversation.call_args,
```
dbajet: Here the assert says `single_conversation` was called with the arguments used to called `single_conversation`, which means it will be always true.

dbajet (changes requested):
Thank you @alexiadowns-canvas for all the modifications. 🙏

This is going in the right direction.

There are still important needed changes:
1. now that all the commands check first, through the method `is_available`, if the fields are editable, is the `resolve_field` method even needed? rather than going directly to `fill_template_content`?
The commands `history_of_present_illness`, `instruct`, `plan` and `reason_for_visit` already are doing it.
One way or another, it should be consistent.

2. Some advertized changes are not present:
 - `if TYPE_CHECKING:` is still present
 - `ImplementedCommands.template_command_list()` is still there
 - every mock in every test has a mock_calls assertion

3. Several tests have the same issues:
    - the mock calls are not checked
    - the test name doesn't follow the convention `test_<method>__<scenario>` (this is not a big deal)
    - some scenarios are missing

4. `TemplatePermissions` design does not offer a singleton. The part `5. Singleton per note_uuid` in the [comment](https://github.com/canvas-medical/canvas-hyperscribe/pull/204/#issuecomment-3894840336) is misleading: with the committed code, the `~33 commands share one instance`... that is cached and retrieved by each command.


5. In `tests/hyperscribe/commands/test_base.py`, when prompts are involved in the source code, there is no actual checks of them. The should be a point where these strings are tested.

hyperscribe/libraries/template_permissions.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+class TemplatePermissions:
+    """Loads note-template edit permissions cached by the note_templates plugin.
+
+    Cache key: ``note_template_cmd_perms_{note_uuid}``
+    Value: ``{CommandType: {plugin_can_edit, field_permissions: [...]}}``
+    """
+
+    def __init__(self, note_uuid: str, cache_getter: Callable[[], Cache] | None = None) -> None:
```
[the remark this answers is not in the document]
alexiadowns-canvas: keeping this comment open for testing!

tests/hyperscribe/commands/test_update_goal.py
```diff
@@ -303,6 +312,19 @@ def reset_mocks():
...
         assert result is expected
+        calls = [call("progress")]
+        assert can_edit_field.mock_calls == calls
         calls = [call()]
         assert current_goals.mock_calls == calls
         reset_mocks()
+
+
+@patch.object(UpdateGoal, "can_edit_field", return_value=False)
+def test_is_available__all_fields_locked(can_edit_field):
```
dbajet: Same issue than with `tests/hyperscribe/commands/test_assess.py::test_is_available__all_fields_locked`: changing to `@patch.object(CloseGoal, "can_edit_field", return_value=True)` will not break the test.

dbajet (changes requested):
@alexiadowns-canvas Things are moving good. :slightly_smiling_face:

There are issues with the tests (related to the commands) as some are incorrect.
What they test is always validated regardless of the changes (e.g. `False And <value>` is always `False`). :disguised_face:

Additionally, when the `result` is compared to `True`, `False` or `None`, the assertion should use `is`, not `==` (i.e. `assert result is expected`)

hyperscribe/commands/base.py
```diff
@@ -13,16 +17,27 @@
 
 
 class Base:
-    def __init__(self, settings: Settings, cache: LimitedCache, identification: IdentificationParameters):
+    def __init__(
+        self,
+        settings: Settings,
+        cache: LimitedCache,
+        identification: IdentificationParameters,
+        permissions: TemplatePermissions | None = None,
```
dbajet: As mentioned in a previous review: using optional parameters is a common practice, widely accepted but to avoid as it obfuscates the code (and it contradicts the Python mantra _be explicit_).

For common parameters, there is always the option to create wrappers:
```python
def function(parm1, parm2, parm3):
   ...

def wrapper_for_common_params(parm1):
   return function(parm1, value2, value3)
```


Additionally in the present case, it is only 'saving' some changes in the tests since the constructor is only called in one place in the code [AudioInterpreter](hyperscribe/libraries/audio_interpreter.py), with the parameter.


The changes to apply in the tests are to add the parameter in the command constructor in the `helper_instance` method, for example in [test_vitals.py](tests/hyperscribe/commands/test_vitals.py):
```python
from hyperscribe.libraries.template_permissions import TemplatePermissions


def helper_instance() -> Vitals:
    ... # no changes
    return Vitals(settings, cache, identification, TemplatePermissions("noteUuid"))

```

tests/hyperscribe/libraries/test_audio_interpreter.py
```diff
@@ -127,22 +131,89 @@ def reset_mocks():
...
     assert instance.cache == cache
-
     calls = [call()]
     assert command_list.mock_calls == calls
     for idx, mock in enumerate(mocks):
+        # With walrus operator optimization, class_name() is called only once
+        # (assigned via := in policy check) and reused for dict key
         calls = [
-            call(settings, cache, identification),
+            call(settings, cache, identification, ANY),
```
dbajet: The `ANY` will go away when the `TemplatePermissions` instance is a  required parameter.
`TemplatePermissions` could be mocked:
```python

@patch("hyperscribe.libraries.audio_interpreter.TemplatePermissions")
@patch.object(ImplementedCommands, "command_list")
def test___init__(command_list, template_permissions):
    mocks = [MagicMock(), MagicMock(), MagicMock(), MagicMock(), MagicMock()]

    def reset_mocks():
        command_list.reset_mocks()
        template_permissions.reset_mocks()
        for item in mocks:
            item.reset_mock()

    [...]
    command_list.side_effect = [mocks]
    template_permissions.side_effect = ["templatePermissionsInstance"]

    [...]
    instance = AudioInterpreter(settings, aws_s3, cache, identification)
    [...]
    calls = [call()]
    assert command_list.mock_calls == calls
    calls = [call("noteUuid")]
    assert template_permissions.mock_calls == calls
    [...]
    reset_mocks()


```

dbajet (changes requested):
@alexiadowns-canvas it is nice to see the extra effort to inject the `TemplatePermission`s instance into the `Base` constructor! :hugs:


But doing it through an optional parameter is easily avoided (even if it means changing all the command tests).

That should be the last changes! :zipper_mouth_face:

Issue #208, opened 2026-02-12
Fix denials for medication statement


alexiadowns-canvas:
The LLM is selecting a medication when the transcript states "no medications"

PR #209, opened 2026-02-13
Prevents medication statement denials from selecting medications

alexiadowns-canvas:
Addresses issue #208 

Feedback: Scribe tried to add a medication statement for "flaxseed oil" which patient is not taking and was not mentioned.

Transcript contained clinician confirming that the patient was NOT taking any medication, which led to a selection of a non-medication

### DIFF
--- a/hyperscribe/commands/medication.py
+++ b/hyperscribe/commands/medication.py
@@ -35,8 +35,8 @@ class Medication(Base):
     ) -> InstructionWithCommand | None:
         result = MedicationStatementCommand(sig=instruction.parameters["sig"], note_uuid=self.identification.note_uuid)
         # retrieve existing medications defined in Canvas Science
-        expressions = instruction.parameters["keywords"].split(",")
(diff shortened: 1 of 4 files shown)


hyperscribe/commands/medication.py
```diff
@@ -35,8 +35,8 @@ def command_from_json(
     ) -> InstructionWithCommand | None:
         result = MedicationStatementCommand(sig=instruction.parameters["sig"], note_uuid=self.identification.note_uuid)
         # retrieve existing medications defined in Canvas Science
-        expressions = instruction.parameters["keywords"].split(",")
-        if medications := CanvasScience.medication_details(expressions):
+        expressions = [e.strip() for e in instruction.parameters["keywords"].split(",") if e.strip()]
```
dbajet: This change is duplicating the one in `hyperscribe/libraries/canvas_science.py`, which is the most important.

Issue #210, opened 2026-02-17
User unable to restart Hyperscribe after commander failure


alexiadowns-canvas:
A user reported "transcript not loading" for note dc30bfeb-7b5d-4f60-bb9f-7b1d204eeed0 (feedback ID 38b4260e). 

Investigation found 17 audio chunks uploaded to S3 between 18:01:11 and 18:05:05 UTC on Feb 16, spaced ~15 seconds apart. Recording worked but no transcripts, no finals, no logs indicating that the commander never ran (no errors logged on either side). 
                                                                                                                                            
The root cause is in capture_view.py:245. Each time _add_cycle receives an audio chunk, it checks stop_and_go.is_running() before launching the commander. If is_running is True, it assumes a commander is already processing and skips. But if a previous commander crashed or the plugin worker was killed mid-execution (OOM, restart), the finally block that sets is_running=False never fires. The flag stays True in cache (14-day TTL), and every subsequent audio chunk silently skips commander launch.

PR #211, opened 2026-02-17
Relaunches commander if there are audio chunks waiting

alexiadowns-canvas:
**Problem**: If a Hyperscribe session's commander crashes, StopAndGo state persists with is_running=True and the finally block never executes. New audio chunks keep arriving via _add_cycle and pile up in waiting_cycles, but the if not stop_and_go.is_running() gate prevents a new commander from launching. The session is silently stuck.                                     
                                                                                                                                            
**Fix**: Auto-recover stuck sessions in _add_cycle. When is_running=True and 5+ waiting cycles have accumulated (~25s of unprocessed chunks), we log a warning with session details and flip is_running to False on the existing StopAndGo (preserving waiting_cycles, cycle, etc.), so the relaunched commander picks up where the dead one left off. The existing new_session_post detection remains as a fallback for cases where chunks stopped flowing entirely (e.g. user closed browser, then comes back later).

Addresses #210

### DIFF
--- a/hyperscribe/handlers/capture_view.py
+++ b/hyperscribe/handlers/capture_view.py
@@ -123,7 +123,6 @@ class CaptureView(SimpleAPI):
         )
 
         stop_and_go = StopAndGo.get(note_id)
-
         context = {
             "patientUuid": patient_id,
             "noteUuid": note_id,
@@ -242,7 +241,6 @@ class CaptureView(SimpleAPI):
             if not response.status_code:
                 return [Response(b"Failed to save chunk (AWS S3 failure)", HTTPStatus.SERVICE_UNAVAILABLE)]
             return [Response(response.content, HTTPStatus(response.status_code))]
-
         if not stop_and_go.is_running():
             user_id = self.request.headers.get("canvas-logged-in-user-id")
             executor.submit(Helper.with_cleanup(self.run_commander), identification, user_id)
--- a/tests/hyperscribe/handlers/test_capture_view.py
+++ b/tests/hyperscribe/handlers/test_capture_view.py
@@ -165,25 +165,21 @@ def test_session_progress_log(progress):
 @patch("hyperscribe.handlers.capture_view.render_to_string")
 @patch("hyperscribe.handlers.capture_view.Authenticator")
 def test_capture_get(authenticator, render_to_string, stop_and_go, helper, customization):
-    mock_stop_and_go = MagicMock()
-
     def reset_mocks():
         authenticator.reset_mock()
         render_to_string.reset_mock()
         stop_and_go.reset_mock()
         helper.reset_mock()
         customization.reset_mock()
-        mock_stop_and_go.reset_mock()
 
     render_to_string.side_effect = ["<html/>"]
     helper.canvas_ws_host.side_effect = ["theWsHost"]
     customization.customizations.side_effect = [Customization(ui_default_tab=DefaultTab.ACTIVITY, custom_prompts=[])]
     authenticator.presigned_url.side_effect = ["Url1"]
     authenticator.presigned_url_no_params.side_effect = ["Url2", "Url3", "Url4", "Url5", "Url6", "Url7", "Url8", "Url9"]
-    mock_stop_and_go.is_ended.side_effect = [False]
-    mock_stop_and_go.is_paused.side_effect = [False, False]
-    mock_stop_and_go.cycle.side_effect = [7]
-    stop_and_go.get.side_effect = [mock_stop_and_go]
+    stop_and_go.get.return_value.is_ended.side_effect = [False]
+    stop_and_go.get.return_value.is_paused.side_effect = [False, False]
+    stop_and_go.get.return_value.cycle.side_effect = [7]
 
     tested = helper_instance()
     tested.request = SimpleNamespace(
@@ -260,15 +256,14 @@ def test_capture_get(authenticator, render_to_string, stop_and_go, helper, custo
         ),
     ]
     assert render_to_string.mock_calls == calls
-    calls = [call.get("the-00-note")]
-    assert stop_and_go.mock_calls == calls
     calls = [
-        call.is_ended(),
-        call.is_paused(),
-        call.cycle(),
-        call.is_paused(),
+        call.get("the-00-note"),
+        call.get().is_ended(),
+        call.get().is_paused(),
+        call.get().cycle(),
+        call.get().is_paused(),
     ]
-    assert mock_stop_and_go.mock_calls == calls
+    assert stop_and_go.mock_calls == calls
     calls = [call.canvas_ws_host("customerIdentifier")]
     assert helper.mock_calls == calls
     calls = [
@@ -292,7 +287,6 @@ def test_new_session_post(session_progress_log):
         path_params={"patient_id": "thePatientId", "note_id": "theNoteId"},
         headers={"canvas-logged-in-user-id": "theUserId"},
     )
-
     result = tested.new_session_post()
     assert result == []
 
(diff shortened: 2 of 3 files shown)


hyperscribe/handlers/capture_view.py
```diff
@@ -157,6 +157,15 @@ def capture_get(self) -> list[Response | Effect]:
     def new_session_post(self) -> list[Response | Effect]:
         patient_id = self.request.path_params["patient_id"]
         note_id = self.request.path_params["note_id"]
+        stop_and_go = StopAndGo.get(note_id)
+        if stop_and_go.is_running() and stop_and_go.waiting_cycles():
```
dbajet: To access this part of the code, the button `Start` has to be enabled and a session already started without being stopped or paused.

It is likely because the page has been re-lauded.
So while, indeed the session might be stuck, it might not be stuck:
- start a session
- add 1 or 2 cycles added
- refresh the page
- click on start

Since the stuck session is managed later, the code here might be more armful than helpful.

Additionally, I would reinforce the logic in regards to the session (i.e. `StopAndGo`) in the `capture_get` method, where the variables are set on a (re)laud of the page.

tests/hyperscribe/handlers/test_capture_view.py
```diff
@@ -277,21 +277,97 @@ def reset_mocks():
...
+    ]
+    assert stop_and_go.mock_calls == calls
+    calls = [call()]
+    assert mock_existing.is_running.mock_calls == calls
+    calls = [call(), call()]
+    assert mock_existing.waiting_cycles.mock_calls == calls
+    calls = [call()]
+    assert mock_existing.cycle.mock_calls == calls
+    calls = [call()]
+    assert mock_existing.created.mock_calls == calls
```
dbajet: The mock itself should be checked against the `mock_calls`, not only a few picked methods:
```python
    calls = [
        call.is_running(),
        call.waiting_cycles(),
        call.cycle(),
        call.waiting_cycles(),
        call.created(),
    ]
    assert mock_existing.mock_calls == calls
```

dbajet (changes requested):
The PR introduces - rightly - a way to recover from a block state.

It seems important to understand why the state is locked.

There are few `try/except` in the code, and the one on top (in `hyperscribe/handlers/capture_view.py::run_commander`) has a `finally` that should set to `False` the `is_running` flag.

Could it be that the "crash" occurs when a new version of the Hyperscribe is pushed? A new version of the Canvas code itself? A LLM call hanging forever? Or the user refreshed the page?

So, while adding a recovery system seems helpful for now, investigating and addressing the root causes of the "crashes" seem necessary.

From a design perspective, moving the logic to identify and fix a stuck session in the `StopAndGo` class would centralize the state management.


In regards to the tests:
- the mock themselves have to be checked using `mock_call`.
- the mock have to be set up using `side_effect` not `return_value`.

alexiadowns-canvas:
> Could it be that the "crash" occurs when a new version of the Hyperscribe is pushed? A new version of the Canvas code itself? A LLM call hanging forever? Or the user refreshed the page?

The issue occurred between 10:01:11 AM and 10:05:05 AM PST on Feb 16th, which didn't coincide with a Hyperscribe deployment or a Canvas code release. I tried to replicate the circumstances by reinstalling Hyperscribe during a session (no crash) and refreshing the page (no crash). I'll keep looking into the LLM calls

hyperscribe/handlers/capture_view.py
alexiadowns-canvas: This is a really good point

alexiadowns-canvas: Removed stuck session detection (lines ~160-168)

tests/hyperscribe/handlers/test_capture_view.py
```diff
@@ -762,6 +838,69 @@ def reset_mocks():
...
+        call(aws_s3_credentials),
+        call().upload_binary_to_s3("theS3Path", b"theContent", "content/type"),
+    ]
+    assert aws_s3.mock_calls == calls
+    calls = [
+        call.warning(
+            "Stuck session detected for note theNoteId: "
+            "cycle=20, "
+            "waiting=[21, 22, 23, 24, 25], "
+            f"created={mock_created.isoformat()}"
```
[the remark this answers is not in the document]
alexiadowns-canvas: that makes sense!

hyperscribe/libraries/stop_and_go.py
```diff
@@ -66,6 +69,16 @@ def add_waiting_cycle(self) -> StopAndGo:
         if self._waiting_cycles:
             next_cycle = self._waiting_cycles[-1] + 1
         self._waiting_cycles.append(next_cycle)
+
```
dbajet: As mentioned above, this soft recovery should be in the `get` method, which means it is called transparently each time the object is retrieved (it is a minimal overhead).

```python
[...]
    def add_waiting_cycle(self) -> StopAndGo:
        next_cycle = self._cycle + 1
        if self._waiting_cycles:
            next_cycle = self._waiting_cycles[-1] + 1
        self._waiting_cycles.append(next_cycle)
        return self._recover_stuck_session()
[...]
  
    def _recover_stuck_session(self)-> StopAndGo:
        if self._is_running and len(self._waiting_cycles) >= Constants.STUCK_SESSION_WAITING_CYCLES_THRESHOLD:
            self._is_running = False
            log.warning(
                f"Stuck session detected for note {self.note_uuid}: "
                f"cycle={self.cycle()}, "
                f"waiting={self.waiting_cycles()}, "
                f"created={self.created().isoformat()}"
            )
        return self
[...]
    @classmethod
    def get(cls, note_uuid: str) -> StopAndGo:
        if cached := get_cache().get(f"stopAndGo:{note_uuid}"):
            return StopAndGo.load_from_json(cached)._recover_stuck_session()
        return StopAndGo(note_uuid)
[...]
```

tests/hyperscribe/libraries/test_stop_and_go.py
```diff
@@ -173,6 +182,61 @@ def test_add_waiting_cycle():
...
+    reset_mocks()
+
+    # below threshold while running: no warning, stays running
+    mock_datetime.now.side_effect = [datetime(2025, 1, 1, 0, 0, 0, tzinfo=UTC)]
+    tested = StopAndGo("theNoteUuid")
+    tested._is_running = True
+    for _ in range(Constants.STUCK_SESSION_WAITING_CYCLES_THRESHOLD - 1):
+        tested.add_waiting_cycle()
+    assert tested.is_running() is True
+    assert len(tested.waiting_cycles()) == Constants.STUCK_SESSION_WAITING_CYCLES_THRESHOLD - 1
```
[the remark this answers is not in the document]
alexiadowns-canvas: Threshold-related tests now use patch.object(Constants, "STUCK_SESSION_WAITING_CYCLES_THRESHOLD", 3) with hardcoded expected values:

tests/hyperscribe/libraries/test_stop_and_go.py:                                                                                     
  - test_add_waiting_cycle: line 190                                                                                                   
  - test__recover_stuck_session: line 249                                                                                              
  - test_get (stuck session recovered on load): line 523

alexiadowns-canvas:
All the logic related to stuck session could be moved to the StopAndGo class, making it fully transparent to CaptureView. - yes, done! ✅ 

For the tests, it is important to use hard coded values (using the constants as input is fine, not as output/expected values). - fixed, thank you ✅

tests/hyperscribe/handlers/test_capture_view.py
dbajet: Since there is no more changes in `hyperscribe/handlers/capture_view.py`, there is no change for the tests (no breaking lines, no additional cases).
All changes here have no impact on the tests and should be removed.

tests/hyperscribe/libraries/test_stop_and_go.py
```diff
@@ -389,8 +516,42 @@ def reset_mocks():
     assert result._delay == 3
     calls = [call(), call().get("stopAndGo:theNoteUuid")]
     assert get_cache.mock_calls == calls
+    assert mock_log.mock_calls == []
     reset_mocks()
 
+    # key exists, stuck session recovered on load
```
dbajet: Mocking directly `_recover_stuck_session` would have simplify the test while covering correctly the logic and code.

tests/hyperscribe/libraries/test_stop_and_go.py
```diff
@@ -158,7 +159,15 @@ def test_reset_paused_effect():
...
-def test_add_waiting_cycle():
+@patch("hyperscribe.libraries.stop_and_go.log")
+@patch("hyperscribe.libraries.stop_and_go.datetime", wraps=datetime)
+def test_add_waiting_cycle(mock_datetime, mock_log):
+    def reset_mocks():
+        mock_datetime.reset_mock()
+        mock_log.reset_mock()
+
+    # basic behavior: not running, below threshold
+    mock_datetime.now.side_effect = [datetime(2025, 1, 1, 0, 0, 0, tzinfo=UTC)]
```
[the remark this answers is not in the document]
alexiadowns-canvas: agreed, perfect date is gone!

PR #212, opened 2026-02-17
Prevent physical exam systems from being disabled

alexiadowns-canvas:
Clinicians have given the feedback that they want to carry forward previous physical exam findings from previous visits in their note. If there is an updated finding during the session, they want Hyperscribe to update the relevant body system. If there are no updated findings, they want Hyperscribe to leave the body system unedited (rather than resetting it and clearing the content).

### DIFF
--- a/hyperscribe/commands/base_questionnaire.py
+++ b/hyperscribe/commands/base_questionnaire.py
@@ -32,20 +32,6 @@ class BaseQuestionnaire(Base):
     def sdk_command(self) -> Type[QuestionnaireCommand]:
         raise NotImplementedError
 
-    def additional_instructions(self) -> list[str]:
-        return []
-
-    def skipped_field_instruction(self) -> str:
-        # The imperative tone is intentional — the LLM must never flip 'skipped' back to True
-        # once a question has been enabled. This is enforced both by this prompt and by
-        # post_process_questionnaire as a structural safeguard.
-        return (
-            "CRITICAL: If a question already has 'skipped' set to 'false', you MUST keep it as 'false'. "
-            "Never change 'skipped' from 'false' back to 'true' or 'null'. "
-            "You may only change 'skipped' from 'true' to 'false' if the question is clearly addressed "
-            "in the transcript. Questions that are already enabled must stay enabled."
-        )
-
     @classmethod
     def staged_command_extract(cls, data: dict) -> CodedItem | None:
         questionnaire = ((data.get("questionnaire") or {}).get("extra")) or {}
@@ -234,7 +220,7 @@ class BaseQuestionnaire(Base):
                 "Your response must be the JSON Markdown block of the questionnaire, with all the necessary "
                 "changes to reflect the transcript content.",
                 "",
-            ] + self.additional_instructions()
+            ]
             transcript = json.dumps([line.to_json() for line in discussion], indent=1)
 
             user_prompt = [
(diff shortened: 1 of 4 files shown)


hyperscribe/commands/physical_exam.py
dbajet: The `additional_instructions` as defined in this file `hyperscribe/commands/physical_exam.py`, in an ealier PR, includes characters (`\u2013` and `\u2014`) that don't help the LLM (they are just dashess) but make the prompt difficult to read for humans.
They should be converted to the ASCII form (`-`)

hyperscribe/commands/physical_exam.py
```diff
@@ -25,5 +37,54 @@ def additional_instructions(self) -> list[str]:
...
+    def post_process_questionnaire(
+        self,
+        original: QuestionnaireDefinition,
+        updated: QuestionnaireDefinition,
+    ) -> QuestionnaireDefinition:
+        """Prevent the LLM from clearing existing PE findings or disabling body systems."""
+        original_by_id = {q.dbid: q for q in original.questions}
+        fixed_questions: list[Question] = []
+
+        for uq in updated.questions:
```
dbajet: While using `q` in the short comprehension expression above (line 46) is helping the readability, using `uq` and `oq` in this long loop is not.
Changing `uq` as `upd_ques`, and `oq` as `new_ques` would help.

And later: `ur` as `upd_resp` and `og_r` as `new_resp`.

Or something more explicit than 2 combinaison of similar letters (uq and oq).

tests/hyperscribe/commands/test_physical_exam.py
```diff
@@ -104,6 +108,245 @@ def test_additional_instructions():
...
+                dbid=11,
+                label="HEENT",
+                type=QuestionType.TYPE_TEXT,
+                skipped=False,
+                responses=[Response(dbid=101, value="updated findings", selected=False, comment=None)],
+            ),
+        ],
+    )
+    result = tested.post_process_questionnaire(original, updated)
+    # General: skipped reverted from True back to False (original was False)
```
dbajet: Using a full `expected` object will prevent any side effect or change on other fields not targeted today:
```python
    expected = Questionnaire(
        dbid=1,
        name="PE",
        questions=[
            # General: skipped reverted from True back to False (original was False)
            Question(
                dbid=10,
                label="General",
                type=QuestionType.TYPE_TEXT,
                skipped=False,
                responses=[Response(dbid=100, value="normal", selected=False, comment=None)],
            ),
            # HEENT: stayed False, text updated normally
            Question(
                dbid=11,
                label="HEENT",
                type=QuestionType.TYPE_TEXT,
                skipped=False,
                responses=[Response(dbid=101, value="updated findings", selected=False, comment=None)],
            ),
        ],
    )
    assert result == expected
```

hyperscribe/commands/base_questionnaire.py
```diff
@@ -31,6 +31,16 @@ def include_skipped(self) -> bool:
...
+    def additional_instructions(self) -> list[str]:
+        return []
+
+    def skipped_field_instruction(self) -> str:
+        """LLM instruction for handling the 'skipped' field. Override for custom semantics."""
+        return (
+            "This includes the values of 'skipped', change it to 'false' only if the question "
+            "is obviously answered in the transcript, don't change it at all otherwise."
+        )
+
```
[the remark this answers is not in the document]
alexiadowns-canvas: BaseQuestionnaire.skipped_field_instruction() now uses the imperative CRITICAL: form, and the PhysicalExam override was removed

hyperscribe/commands/physical_exam.py
alexiadowns-canvas: both \u2013 (–) and \u2014 (—) were replaced with ASCII dashes (-) in additional_instructions.

dbajet (approved):
@alexiadowns-canvas this looks great! :100:

The post-process is definitively an improvement.

Users may need to be aware that they cannot vocalize skipping an enabled question.

PR #221, opened 2026-03-02
Validate LLM index against name for conditions and goals

alexiadowns-canvas:
When the LLM identifies a condition or goal from a list, it returns both a name (e.g., "Generalized edema") and an index (e.g., position 9). The name is almost always correct, but with long lists it's possible for the LLM to miscount the index. In these cases, an error that picks the wrong index by one (9 vs 10) means picking the wrong condition entirely (e.g., "Dry eye syndrome" instead of "Generalized edema").                                                                 
   
The fix checks if the index and name agree. If they don't, it falls back to the name, which is the more reliable signal. 

This applies to all 6 command types that do index-based lookups (Assess, ResolveCondition, Prescription, UpdateDiagnose, CloseGoal, UpdateGoal).

Addresses issue #216

### DIFF
--- a/hyperscribe/commands/assess.py
+++ b/hyperscribe/commands/assess.py
@@ -34,9 +34,13 @@ class Assess(Base):
         chatter: LlmBase,
     ) -> InstructionWithCommand | None:
         condition_id: str | None = None
-        if 0 <= (idx := instruction.parameters["conditionIndex"]) < len(current := self.cache.current_conditions()):
-            condition_id = current[idx].uuid
-            self.add_code2description(current[idx].uuid, current[idx].label)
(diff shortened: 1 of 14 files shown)


beaugunderson (approved):

beaugunderson:
i'm curious if the LLM ever slightly changes the name, e.g. does it make sense to allow for levenshtein distance of 1 or 2 in the comparison? (don't have to do it in this PR, but if we see the name-based lookup fail often it would be worth looking at)

alexiadowns-canvas:
@beaugunderson interesting and now we'll know...
added https://github.com/canvas-medical/canvas-hyperscribe/issues/222#issue-4015126144

PR #225, opened 2026-03-10
Buttons in the plan section for Add Task and Add Order

aduane:
Task works

Order - Rx mostly works
Order - Imaging not started
Order - Lab not started

### DIFF
--- a/hyperscribe/scribe/api/session_view.py
+++ b/hyperscribe/scribe/api/session_view.py
@@ -13,6 +13,8 @@ from canvas_sdk.handlers.simple_api import SessionCredentials, SimpleAPI, StaffS
 
 from canvas_sdk.v1.data.medication import Medication, MedicationCoding, Status
 from canvas_sdk.v1.data.note import Note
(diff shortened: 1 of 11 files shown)


hyperscribe/scribe/api/session_view.py
```diff
@@ -318,3 +320,18 @@ def get_search_medications(self) -> list[Union[Response, Effect]]:
...
+    @api.get("/assignees")
+    def get_assignees(self) -> list[Union[Response, Effect]]:
+        """Return a merged list of active staff members and teams for task assignment."""
+        assignees: list[dict[str, Any]] = []
+        for s in (
+            Staff.objects.filter(active=True)
+            .order_by("last_name", "first_name")
+            .values("dbid", "first_name", "last_name")
+        ):
+            label = f"{s['first_name']} {s['last_name']}".strip()
```
ruipbatista, line 333: This should probably use `staff.credentialed_name` instead

ruipbatista (approved):

PR #232, opened 2026-03-19
Get mapped ICD-10 when available, then fuzzy match

alexiadowns-canvas:
Made two adjustments to address customer reported condition mismatches:

1) generate-normalized-data endpoint has an include_corresponding_note_problems flag, which is set to false by default. I flipped it to "true" so we can now use their mapping when available to match Assessment & Plan headers to the related condition. If it's not available, we fallback to the original fuzzy match

2) The reported mismatches were using "unspecified" to meet the 0.5 overlap threshold, so I also added some common causes of mismatch to the stop list


nmpsilva (approved):
LGTM! Did a small fix to make the comparison case-insensitive, but that was it.

PR #238, opened 2026-03-25
fix: remove unauthorized module assignment

nmpsilva:
This pull request makes a minor update to the `get_attempts` method in `canvas_science.py`. The change removes the lines that set the `_MAX_REQUEST_TIMEOUT_SECONDS` attribute for both `ontologies_http` and `science_http` before making HTTP requests. This simplifies the method and relies on the default timeout settings for these HTTP clients.

- Removed manual setting of `_MAX_REQUEST_TIMEOUT_SECONDS` for `ontologies_http` and `science_http` in the `get_attempts` method, defaulting to the clients' existing timeout configuration.


alexiadowns-canvas (approved):
Thanks for the update, makes sense!

PR #248, opened 2026-04-10
Fix JWT token expiry and recording state regressions

beaugunderson:
## Summary
- During long Nabla transcription sessions the user access token expires, causing every WebSocket reconnect to fail with error 50025 ("Invalid JWT token: The Token has expired")
- Added `_reconnect()` to `NablaScribeClient` that fetches fresh tokens via a `refreshConfig` callback before each reconnect attempt
- Suppress error 50025 from user display (same pattern as existing 83011 stream-timeout suppression) since reconnection handles it transparently
- **Fix "Paused" badge after finish** — skip `sendBeacon` on unmount when the transcript is already finalized, preventing the backend from overwriting `finalized=true` back to `false` ([KOALA-4818](https://canvasmedical.atlassian.net/browse/KOALA-4818))
- **Fix stuck "Finalizing Transcript..."** — use `forceEnd()` instead of `end()` when finishing recording so the UI doesn't block on buffer drain; reduce `END_TIMEOUT_MS` from 60s to 30s for the pause path ([KOALA-4818](https://canvasmedical.atlassian.net/browse/KOALA-4818))

## Test plan
- [ ] Start a recording session and let it run long enough for the Nabla JWT to expire (or simulate by shortening the token TTL)
- [ ] Verify the WebSocket reconnects silently with fresh tokens — no red error text should appear
- [ ] Verify normal pause/resume and network-loss reconnection still work
- [ ] Pause → Finish → Accept & Prescribe → return to note: should **not** show "Paused" badge
- [ ] Start → Finish (no pause): "Finalizing Transcript..." should resolve quickly, not hang for 60s

[KOALA-4818]: https://canvasmedical.atlassian.net/browse/KOALA-4818?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[KOALA-4818]: https://canvasmedical.atlassian.net/browse/KOALA-4818?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/hyperscribe/scribe/static/recording-hook.js
+++ b/hyperscribe/scribe/static/recording-hook.js
@@ -84,15 +84,25 @@ export function useRecording(noteId, initialTranscript) {
       return false;
     }
 
+    const refreshConfig = async () => {
+      const res = await fetch(`${API_BASE}/config`, { cache: 'no-store' });
+      return res.json();
+    };
+
     let client;
     try {
-      client = createScribeClient(config);
+      client = createScribeClient({ ...config, refreshConfig });
       client.onTranscriptItem = handleTranscriptItem;
       client.onError = (msg, code) => {
         if (code === 83011) {
           // Stream timeout — Nabla closes the connection if a stream receives
           // no audio for 10s. Not actionable for the user; the client will
           // reconnect and replay buffered audio automatically.
+          return;
+        }
+        if (code === 50025) {
+          // JWT expired — the client will refresh the token and reconnect
+          // automatically. Not actionable for the user.
           return;
         }
         setError(code ? `${msg} (${code})` : msg);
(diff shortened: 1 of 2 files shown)


aduane (approved):

nmpsilva (approved):

kevincareycanvas:
Tested this and encountered no issues. Tested with a transcript 80 minutes long.

beaugunderson:
@kevincareycanvas beautiful! merging

PR #251, opened 2026-04-16
Pull mental health exam section for psychiatry ROS

alexiadowns-canvas:
1) Selects `PSYCHIATRY_MULTIPLE_SECTIONS` Nabla template when provider chooses the Psychiatry visit template
2) Adds `MENTAL_HEALTH_EXAM` custom instruction with psychiatric ROS categories (Depressive Symptoms, Anxiety Symptoms, Sleep, Appetite, SI/HI, Hallucinations, Delusions/Paranoia, Manic Symptoms)
3)Maps `mental_health_exam` to ROS command in the extractor, preferred over generic review_of_systems when present
4) Merge separate Assessment and Plan sections into assessment_and_plan for downstream extractor and ap_split compatibility, assessment items become header-only blocks for ICD-10 matching, plan items become header+body blocks with plan details preserved
5) Reformat Plan bullets into header+body blocks so parse_ap_blocks() can match conditions to ICD-10 codes because the psychiatry template does not support `split_by_problem`


nmpsilva (commented):
If we're moving forward with the Psychology template, this looks good. 
However, I need to point out a couple of things:
1. No tests: we need those, we like tests :)
2. The AP merge is unconditional and will happen whenever both 'assessment' and 'plan' keys exist. Maybe not a bug, but maybe we want to limit this to the Psychology template?
3. Looks like the template matching is too sensitive? 'psychiatry' or 'psychiatry visit' won't match will they?
4. The word overlap is confusing. I'm honestly not sure exactly what it's doing. So are we really sure we want to add this complexity? (The test will help, but the question stands). Still related to exactly this, Claude did point out a potential issue: 
    
> When an assessment item has zero word overlap with any plan block header (best_score stays 0.0), it defaults to best_idx = 0 — the first plan block. Unrelated assessment items silently pile up under the first diagnosis. A safer default would be to leave unmatched items as standalone header-only blocks.

alexiadowns-canvas:
@claude review

alexiadowns-canvas:
let's go for one more @claude review

alexiadowns-canvas:
Decisions from 5/29:

No tests: we need those, we like tests :) 

- Makes sense... thank you for adding tests 🙏 

The AP merge is unconditional and will happen whenever both 'assessment' and 'plan' keys exist. Maybe not a bug, but maybe we want to limit this to the Psychology template?

- Approach .1: call both templates, and surgically extract the psych template (trade-off is latency)
- Approach .2: just change everything (more complicated, but faster)

Looks like the template matching is too sensitive? 
- From Kevin: we should only select the PSYCHIATRY_MULTIPLE_SECTIONS Nabla template when provider chooses the Psychiatry visit template in Scribe (behavior should not change if provider selects 'Subsequent Visit')
- We need to consider the custom prompt we are using for the HPI when the other visit templates are selected to make sure the ROS matches the template used (per Kevin, ideally we bring in the Psych HPI when the provider chooses the Psychiatry template)

kevincareycanvas:
cerberus review

kevincareycanvas:
## Summary of changes

This branch was **rebased onto the latest `feat/canvas-scribe`** (history rewritten / force-pushed) and then extended with the work below. The full `tests/hyperscribe/scribe` unit suite is green, and both psychiatry and generic visits were validated end-to-end against a live Nabla instance.

### 1. Mental Status Exam section (headline feature)
- New **psychiatry-only Mental Status Exam card** in the scribe summary, rendered directly above Physical Exam and gated to the `Psychiatry` visit template.
- Sources Nabla's `MENTAL_HEALTH_EXAM` section into a new `mental_status_exam` command; mirrors the Physical Exam card (same actions: Clear / Add / Remove / Carry forward / Template).
- The result is written to the NoteBody as a `CustomCommand` with schema_key **`mentalStatusExam`**, declared in `CANVAS_MANIFEST.json` so the instance registers it.
- `_extract_ros` reverted to baseline (generic ROS unchanged); psychiatry gating is surfaced per-template via an `is_psychiatry` flag on `/visit-templates`.
- Carry-forward (`/last-exam`), print/display mappings, and optional template `mse_sections` plumbing added.

### 2. Nabla API migration to `2026-06-12`
- Psychiatry now uses **`PSYCHIATRY_MULTIPLE_SECTIONS_AP_MERGED`** with a single, problem-structured `ASSESSMENT_AND_PLAN` (`split_by_problem`), so the A&P parses through the **same path as generic visits** — no local re-merge on the happy path.
- Bumped `_NABLA_API_VERSION` → `2026-06-12`; renamed request fields `note_template` → `note_template_key` and `template` → `template_key`.
- **Fixed the REST version header name to `X-Nabla-Api-Version`** (was `nabla-api-version`, which is the query-param spelling — Nabla ignored it as a header and silently pinned REST calls to the org default `2026-02-20`). This was the root cause of psychiatry silently falling back to a generic note and `generate-normalized-data` returning 400 under the renamed fields.
- Added `NablaClient.get_note_template()` + `NablaBackend.supports_psychiatry_ap_split_by_problem()` capability check (diagnostics/UAT).

### 3. A&P parsing
- Reworked `_reformat_plan_as_ap` (now only the **dormant fallback** for separate-section responses): assessment items become the diagnosis headers, and plan actions attach to the closest diagnosis (cross-cutting actions default to the primary/first).
- Strip a trailing colon from A&P block headers so `split_by_problem` "Diagnosis:" headers exact-match Nabla's colon-less `corresponding_note_problem` (fixed e.g. "Right rotator cuff tendinitis" missing its M75.41 code).

### 4. Prompts
- Psychiatric ROS appended to the psych HPI; a fixed 11-category Mental Status Exam instruction; US spelling "Judgment" aligned across prompt + docs + the visit-template scaffold.

### Deployment / ops notes (config, not code)
- The `Psychiatry` entry in the `VisitTemplates` secret needs an `mse_template` scaffold to populate the MSE card's Template button (now documented in `hyperscribe/README.md`).
- The instance must have a `mentalStatusExam` custom command registered (declared in the manifest; registers on install/deploy).

### Validation
- Full `tests/hyperscribe/scribe` suite green; lint clean.
- Live UAT: psychiatry note → AP-merged template, problem-structured A&P with correct ICD-10s, full 11-category MSE; generic visit unchanged; normalized data / vitals / conditions working on `2026-06-12`.

kevincareycanvas:
cerberus review

PR #254, opened 2026-04-30
Scribe: read-only mode, locked-note handling, transcript polish

kevincareycanvas:
## Summary

Three buckets of work, shipped together:

**1. Read-only mode + locked-note handling** (the original `changes.md` ask)
- Only the note author (`note.provider.id == logged-in staff_id`) can edit the Scribe tab. Non-authors see a dimmed render with an info-icon banner and no edit / Accept-and-Sign affordances.
- Locked notes drop the lock-icon "Note Locked" placeholder; content is rendered in the same dimmed state with a padlock banner (no editing for anyone). Lock takes precedence over the non-author message.
- `ScribeApp.open_by_default` returns False when a note is locked **and** the Scribe tab has no meaningful content (manual-mode-aware: checks transcript items, summary commands, approved flag, and any non-empty section text). The tab is still in the bar — Canvas just defaults to the legacy note tab.
- Server-side `_authorize_edit` helper guards the six mutating endpoints (`/save-transcript`, `/save-summary`, `/generate-summary`, `/insert-commands`, `/insert-metadata`, `/sign-note`) with 400/403/404. `/insert-metadata` now requires `note_uuid` in the body.

**2. Transcript reliability & polish**
- Force-finalize partial entries on pause / finish so the "Listening" label never persists past live capture; safety net for whatever Nabla doesn't resolve in time.
- Stale-partial promotion: any partial that gets a later final placed after it is force-finalized so it stops rendering as "Listening".
- Monotonic timeline across pause/resume: track `sessionOffsetMs` recomputed on every reconnect; shift incoming `start_offset_ms` / `end_offset_ms` accordingly. Eliminates the bug where post-resume entries interleaved into pre-pause history because Nabla restarts offsets at zero.
- Empty `item_id` partials get a stable synthesized id (no data loss, no stacking).
- End-of-stream signal: stop audio capture and call `client.end()` *before* the speaker-attribution wait so Nabla flushes trailing speaker tags over the still-open WebSocket. `isResolved` check now mirrors the UI's Provider/Patient detection exactly.
- `SPEAKER_WAIT` audit event (reason / resolution / elapsed / unresolved counts) for tuning `FINISH_WAIT_MS` and `PAUSE_WAIT_MS` from real sessions.
- Suppress "Recording in progress" / "Paused" labels once `noteData` or `recording.finalized` is true; the panel header reads "Transcript" instead.
- Fold the "Finalizing transcript" wait into the existing generation progress banner as step 1 of 6 — no more dead time before the LLM kicks in.
- Auto-scroll the transcript to the latest entry while live (RAF-deferred, smooth animation, always pins during recording).

**3. Accept and Sign gating**
- Footer is only visible when **manual mode is active**, or when **AI mode has completed the full record → finish → generate flow** (`mode === 'ai' && recording.finalized && noteData`). Removes the `insertableCount > 0` clause that allowed the button to appear before the user committed to a flow.

## Tests

- New `tests/hyperscribe/scribe/api/test_session_view_auth.py` covering every branch of `_authorize_edit` (missing UUID, note-not-found, not-editable, no provider, different provider, missing staff_id, success).
- New tests in `test_transcript_app.py` for `_scribe_tab_has_content` (no rows, transcript items, approved, commands, section text, blank manual-mode opener, whitespace-only) and `ScribeApp.open_by_default` (editable / locked-with-content / locked-blank / no-note-id).
- Existing endpoint tests get an autouse fixture that bypasses `_authorize_edit`. Two `/insert-metadata` tests updated to pass `note_uuid` in their bodies.
- Backend suite: `uv run pytest tests/hyperscribe -q` → **1491 passed**.

## Test plan

- [ ] Open the Scribe tab as the note author on an editable note: full edit, no banner.
- [ ] Open as a different staffer on the same note: dimmed body, info-icon banner ("Read-only — only the note author can edit the Scribe tab"), no top-bar controls, no Accept and Sign.
- [ ] Lock the note (sign or amend-revert it): body still shows, dimmed, padlock banner ("This note is locked. To make changes, amend the note first.").
- [ ] Lock a never-used Scribe tab: tab still in the bar but Canvas opens the legacy note tab by default. Click into Scribe → empty + locked banner.
- [ ] Lock a note where Scribe was approved: Scribe tab opens by default with the post-approve view.
- [ ] AI flow: Start AI Scribe → speak → click Finish → progress banner shows "Finalizing transcript… (1/6)" through "Suggesting diagnoses… (6/6)" → Accept and Sign appears.
- [ ] Verify "Listening" never persists past Pause or Finish; verify post-resume content lands chronologically below pre-pause content; verify panel header says "Transcript" once `noteData` exists.
- [ ] Auto-scroll: while recording, transcript pins to the bottom even after speaker-attribution flips.
- [ ] Manual mode: click Manual → Accept and Sign appears immediately; AI top controls hidden.
- [ ] curl one of the gated endpoints from a non-author session → 403.

## QA item to confirm in deploy

- Canvas's NoteApplication framework, when no plugin tab returns `open_by_default=True`, falls through to the legacy Canvas note tab (vs. landing on a blank state). This is the assumption behind the locked-blank tab fallback. Easy to revert if not.

### DIFF
--- a/hyperscribe/scribe/api/session_view.py
+++ b/hyperscribe/scribe/api/session_view.py
@@ -33,6 +33,7 @@ from canvas_sdk.v1.data.patient import Patient
 
 from hyperscribe.libraries.canvas_science import CanvasScience
 from hyperscribe.libraries.constants import Constants
+from hyperscribe.libraries.helper import Helper
 
 import hyperscribe.scribe.clients.nabla  # noqa: F401 — register backends
 from hyperscribe.scribe.backend import (
@@ -68,6 +69,31 @@ def _format_icd10_code(raw: str) -> str:
     if len(code) > 3:
         return code[:3] + "." + code[3:]
     return code
+
+
+def _authorize_edit(note_uuid: str, request: Any) -> JSONResponse | None:
+    """Return a JSONResponse to short-circuit when the staff cannot edit the note's Scribe tab.
+
+    Authorized when the note exists, is in an editable state, and the logged-in
+    staff (from `canvas-logged-in-user-id` request header) matches the note's
+    provider. Returns None on success.
+    """
+    if not note_uuid:
+        return JSONResponse({"error": "note_uuid is required"}, status_code=HTTPStatus.BAD_REQUEST)
+    try:
+        note = Note.objects.select_related("provider").get(id=note_uuid)
+    except Note.DoesNotExist:
+        return JSONResponse({"error": "Note not found"}, status_code=HTTPStatus.NOT_FOUND)
+    if not Helper.editable_note(note.dbid):
+        return JSONResponse({"error": "Note is not editable"}, status_code=HTTPStatus.FORBIDDEN)
+    headers = getattr(request, "headers", {}) or {}
+    staff_id = headers.get("canvas-logged-in-user-id") or ""
+    if not staff_id or not note.provider or str(note.provider.id) != str(staff_id):
+        return JSONResponse(
+            {"error": "Only the note author can modify the Scribe tab"},
+            status_code=HTTPStatus.FORBIDDEN,
+        )
+    return None
 
 
 def audit_event(note_uuid: str, event_type: str, details: dict[str, Any] | None = None) -> None:
@@ -516,9 +542,11 @@ class ScribeSessionView(StaffSessionAuthMixin, SimpleAPI):
         note_id = str(data.get("note_id", ""))
         if not note_id:
             return [JSONResponse({"error": "note_id is required"}, status_code=HTTPStatus.BAD_REQUEST)]
+        if denial := _authorize_edit(note_id, self.request):
+            return [denial]
+        staff_id = self.request.headers.get("canvas-logged-in-user-id") or ""
         items = data.get("transcript", {}).get("items", [])
         finalized = bool(data.get("finalized", False))
-        staff_id = self.request.headers.get("canvas-logged-in-user-id") or ""
         _save_transcript(note_id, items, finalized=finalized, provider_id=staff_id)
         return [JSONResponse({"status": "ok"}, status_code=HTTPStatus.OK)]
 
@@ -541,6 +569,8 @@ class ScribeSessionView(StaffSessionAuthMixin, SimpleAPI):
         note_id = str(data.get("note_id", ""))
         if not note_id:
             return [JSONResponse({"error": "note_id is required"}, status_code=HTTPStatus.BAD_REQUEST)]
+        if denial := _authorize_edit(note_id, self.request):
+            return [denial]
         payload: dict[str, Any] = {
             "note": data.get("note", {}),
             "commands": data.get("commands", []),
@@ -593,6 +623,8 @@ class ScribeSessionView(StaffSessionAuthMixin, SimpleAPI):
         note_id = str(data.get("note_id", ""))
         if not note_id:
             return [JSONResponse({"error": "note_id is required"}, status_code=HTTPStatus.BAD_REQUEST)]
+        if denial := _authorize_edit(note_id, self.request):
+            return [denial]
 
         total = len(SUMMARY_STEPS)
 
@@ -1057,6 +1089,8 @@ class ScribeSessionView(StaffSessionAuthMixin, SimpleAPI):
         note_uuid = str(data.get("note_uuid", ""))
         if not note_uuid:
             return [JSONResponse({"error": "note_uuid is required"}, status_code=HTTPStatus.BAD_REQUEST)]
+        if denial := _authorize_edit(note_uuid, self.request):
+            return [denial]
         commands = data.get("commands", [])
         validation_errors = validate_proposals(commands)
         if validation_errors:
@@ -1104,6 +1138,11 @@ class ScribeSessionView(StaffSessionAuthMixin, SimpleAPI):
             data: dict[str, Any] = json.loads(self.request.body)
         except (json.JSONDecodeError, ValueError) as exc:
             return [JSONResponse({"error": f"Invalid JSON: {exc}"}, status_code=HTTPStatus.BAD_REQUEST)]
+        note_uuid = str(data.get("note_uuid", ""))
+        if not note_uuid:
+            return [JSONResponse({"error": "note_uuid is required"}, status_code=HTTPStatus.BAD_REQUEST)]
+        if denial := _authorize_edit(note_uuid, self.request):
+            return [denial]
         pending = data.get("pending", [])
         if not pending:
             return [JSONResponse({"ok": True}, status_code=HTTPStatus.OK)]
@@ -1172,6 +1211,8 @@ class ScribeSessionView(StaffSessionAuthMixin, SimpleAPI):
         note_uuid = str(data.get("note_uuid", ""))
         if not note_uuid:
             return [JSONResponse({"error": "note_uuid is required"}, status_code=HTTPStatus.BAD_REQUEST)]
+        if denial := _authorize_edit(note_uuid, self.request):
+            return [denial]
         audit_event(note_uuid, "SIGN_NOTE", {})
         note_effect = NoteEffect(instance_id=note_uuid)
         return [JSONResponse({"ok": True}, status_code=HTTPStatus.OK), note_effect.lock(), note_effect.sign()]
--- a/hyperscribe/scribe/api/session_view.py
+++ b/hyperscribe/scribe/api/session_view.py
@@ -77,18 +77,22 @@ def _authorize_edit(note_uuid: str, request: Any) -> JSONResponse | None:
     Authorized when the note exists, is in an editable state, and the logged-in
     staff (from `canvas-logged-in-user-id` request header) matches the note's
     provider. Returns None on success.
+
+    Uses `.values()` to fetch only the two scalars we need — this runs on every
+    mutating endpoint, so we want it cheap.
     """
     if not note_uuid:
         return JSONResponse({"error": "note_uuid is required"}, status_code=HTTPStatus.BAD_REQUEST)
     try:
-        note = Note.objects.select_related("provider").get(id=note_uuid)
+        note = Note.objects.values("dbid", "provider__id").get(id=note_uuid)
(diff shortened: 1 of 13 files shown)


hyperscribe/scribe/api/session_view.py
```diff
@@ -70,6 +71,31 @@ def _format_icd10_code(raw: str) -> str:
...
+    """Return a JSONResponse to short-circuit when the staff cannot edit the note's Scribe tab.
+
+    Authorized when the note exists, is in an editable state, and the logged-in
+    staff (from `canvas-logged-in-user-id` request header) matches the note's
+    provider. Returns None on success.
+    """
+    if not note_uuid:
+        return JSONResponse({"error": "note_uuid is required"}, status_code=HTTPStatus.BAD_REQUEST)
+    try:
+        note = Note.objects.select_related("provider").get(id=note_uuid)
```
nmpsilva: Minor perf nit: this fetches the entire Note row (and the entire provider row via select_related) on every gated request, but only `note.dbid` and `note.provider.id` are read. Could be tightened to:

      Note.objects.select_related("provider").values("dbid", "provider__id").get(id=note_uuid)

  Not a blocker, but `_authorize_edit` runs on every mutating endpoint, so it's worth doing.

hyperscribe/scribe/static/recording-hook.js
```diff
@@ -58,19 +117,115 @@ export function useRecording(noteId, initialTranscript) {
...
+  // Nabla restarts start_offset_ms from 0 on every (re)connection, so a fresh
+  // post-resume partial would interleave into the middle of pre-pause content
+  // when sorted chronologically. We add a per-session base — recalculated at
+  // connect time from the current entries' max end offset — so the displayed
+  // offsets stay monotonically increasing across pause/resume cycles. New ids
+  // synthesized via normalizeEntry inherit the session-shifted start so they
+  // remain distinct between sessions.
+  const sessionOffsetMsRef = useRef(0);
+
   const handleTranscriptItem = useCallback((item) => {
```
nmpsilva: I think there's a pause-resume race in this handler that the PR doesn't address. `handleTranscriptItem` blindly applies whatever `sessionOffsetMsRef.current` is at the moment the event arrives, with no gate on the active client/session.

  Repro to try (please verify before merging):

  1. Start recording. Speak ~5 seconds of audio.
  2. Hit Pause. While `waitForSpeakerAttribution` is still running (it polls for ~3s after `client.end()` to drain trailing items), immediately hit Resume.
  3. Continue speaking for another 5+ seconds.
  4. Inspect the rendered transcript timeline order.

  If you see items from before the pause appearing AFTER items from after the resume (or with offsets that look ~5s too high), the race is real: the late-drain events from the old WebSocket got the new `sessionOffsetMsRef.current` applied to them.

hyperscribe/scribe/static/recording-hook.js
```diff
@@ -58,19 +117,115 @@ export function useRecording(noteId, initialTranscript) {
...
+    if (clientRef.current) {
+      clientRef.current.end().catch(() => {});
+    }
+  }, []);
+
   const connectAndRecord = useCallback(async () => {
+    // Capture the session base offset before any new transcript items arrive.
+    // First start: entries is empty → 0. Resume: max end_offset of what's
+    // already there, so new offsets continue monotonically from that point.
+    sessionOffsetMsRef.current = entriesRef.current.reduce(
```
nmpsilva: I've asked Claude to look at this bit specifically, are there may be a bug on long sessions. 
The re-auth path takes a different code route than this one — meaning the offset isn't recalculated when it kicks in. If that's the case, anything spoken AFTER the re-auth could land in the transcript at the wrong place: either jumbled in with stuff from earlier in the visit, or showing weird timestamps

Here's what you can do to test it:


  1. Force the re-auth to fire fast (shorten the token lifetime locally, or just leave a session running for an hour).
  2. Speak a clear marker before the re-auth (e.g. "this is segment one").
  3. Wait for the re-auth.
  4. Speak another clear marker ("this is segment two, after the reconnect").
  5. Generate the summary / open the saved transcript and confirm segment-two text appears AFTER segment-one text, with sensible timestamps — not interleaved or jumping backwards.

If it looks fine, ignore me. If segment two shows up in the middle of segment one (or earlier than it), then there may be a problem

hyperscribe/scribe/api/session_view.py
kevincareycanvas: Great catch — shipped in 3b24842e. Switched to `Note.objects.values('dbid', 'provider__id').get(id=note_uuid)` and dropped the now-redundant `select_related` (the `provider__id` underscore syntax does the join implicitly with `.values()`). Auth-helper test mocks updated to match the new chain.

hyperscribe/scribe/static/recording-hook.js
kevincareycanvas: Great catch — you're right, this is a real race. Shipped a fix in 19a72c02.

Walking through your repro to confirm:
1. T=0: Pause clicked. `signalEndOfStream()` calls `client.end()` (fire-and-forget). The wait begins; old client still alive, draining trailing items.
2. T=0.5s: Nabla pushes a trailing item (raw offset 4500ms). `handleTranscriptItem` reads `sessionOffsetMsRef.current = 0`, adjusted = 4500. Correct.
3. T=1.0s: User hits Resume. `connectAndRecord` recomputes `sessionOffsetMsRef.current = 4500` (max of existing). Opens session B.
4. T=1.5s: Nabla emits *another* trailing item from session A's drain (raw 4800ms). Handler now reads `sessionOffsetMsRef.current = 4500`, adjusted = 9300. Wrong.

The fix captures `sessionOffsetMs` as a local `const` inside `connectAndRecord` and binds it into `client.onTranscriptItem`'s closure:

```js
const sessionOffsetMs = entriesRef.current.reduce(...);
client.onTranscriptItem = (item) => handleTranscriptItem(item, sessionOffsetMs);
```

`handleTranscriptItem` now takes the offset as an arg instead of reading the shared ref. Each scribeClient instance carries its own offset in its handler closure — late drain events from the old client keep applying the old offset, fresh events from the new client get the new offset. Race goes away.

The shared ref is still updated by `connectAndRecord` for compatibility with any future code that needs a "current session" view, but it's no longer read by the handler.

hyperscribe/scribe/static/recording-hook.js
kevincareycanvas: Great catch — looked at this carefully and I think the fix from comment #2 (19a72c02) covers it for free, but worth walking through the analysis.

The re-auth path in `scribeClient.js:_reconnect()`:

```js
async _reconnect() {
  if (this._refreshConfig) {
    const freshConfig = await this._refreshConfig();
    if (freshConfig && freshConfig.access_token) {
      this._config.access_token = freshConfig.access_token;
    }
  }
  this._createWebSocket();
}
```

Two important details:

1. **Same client instance.** `_reconnect` only swaps the access token and calls `_createWebSocket()`. `clientRef.current` doesn't change, so `client.onTranscriptItem` (with its captured `sessionOffsetMs` closure from the original `connectAndRecord`) stays bound. Through any number of internal reconnects, the same offset is applied. We never go back through `connectAndRecord`, so there's no opportunity for the offset to drift mid-session.

2. **Stream continuation, not new session.** `_markSentAsUnsent` (`scribeClient.js:303`) re-queues unacknowledged audio chunks under their original `stream_id` and `seq_id`. That's strong evidence Nabla's protocol treats reconnects as continuations of the existing stream, which means `start_offset_ms` continues from where it left off rather than restarting at zero. Even if Nabla *did* restart the clock for some edge case, the per-client closure would still apply a single consistent offset to everything it receives — items would land in the right relative order, just possibly with a small visual gap.

So the fix in 19a72c02 should make this safe. The only way it could still misbehave is if a single re-auth somehow tore down the entire client and created a new instance via `connectAndRecord`, which doesn't happen in the current code.

Worth running your empirical test in QA anyway:

1. Force the re-auth (shorten token lifetime locally or leave a session running).
2. Speak a clear marker before the re-auth.
3. Wait for re-auth.
4. Speak another marker after.
5. Confirm segment-two text shows up after segment-one with sensible timestamps.

Will mark resolved if you confirm the test looks clean — happy to dig further if anything looks off.

hyperscribe/scribe/static/recording-hook.js
kevincareycanvas: UAT testing passed

hyperscribe/scribe/static/recording-hook.js
kevincareycanvas: Update: ran your empirical test (with a temporary 30s forced-reconnect dev hack to compress an hour of waiting into a few minutes) and **you were right** — my earlier reply here was wrong. The fix in 19a72c02 only handled user-driven reconnects (pause/resume); scribeClient's internal `_reconnect()` reuses the same client instance and thus the same captured-const offset, so post-reconnect items piled up at 0:00 in random order. Repro screenshot showed 4 segments at the same `0:00` timestamp with the order scrambled.

Just shipped 2988e7eb to actually fix it. Two changes:

1. **Mutable session-offset holder.** The handler closure now reads `holder.current` at delivery time instead of capturing a const at bind time. `client.onReconnect` calls a new `refreshSessionOffset()` that updates the holder. The next item arriving through this same client instance picks up the new value, so trailing items from the prior Nabla session keep the prior offset while items from the new session use the new one.
2. **Wall-clock anchoring.** Instead of bumping the offset to `max(end_offset_ms)` (audio-time-without-gaps), the offset is now `Date.now() - recordingStartedAt`. The displayed timeline tracks real visit duration rather than collapsing reconnect dead air. The recording-start anchor is back-dated by saved-transcript `max(end_offset)` on hydration so resumed sessions continue cleanly.

Re-ran the same test with the fix in place: 4 segments spoken across a 2-minute recording with forced reconnects every 30s now display at 0:00, 0:32, 1:01, 1:32 — correct order, monotonic, matching wall-clock to within the reconnect handshake delay.

Sorry for the dismissive earlier reply — your concern was real and the empirical test caught it. Thanks for pushing back.

PR #256, opened 2026-05-02
Switch feedback from Notion to Pylon

alexiadowns-canvas:
This will allow us to reply to customers directly through Pylon when they submit feedback as well as assign and triage tickets dynamically


beaugunderson:
some things worth feeding back into `claude`:

**🔴 Inline import violates project convention** — `hyperscribe/handlers/capture_view.py:328`
```python
from canvas_sdk.v1.data.staff import Staff
```
Move this to the module-level imports. Inline imports are disallowed by project conventions, and this one sits inside `try`/`except` so it re-executes on every feedback POST.

**🟡 Stale failure-path assertions removed** — `tests/hyperscribe/handlers/test_capture_view.py`
The Pylon-failure test no longer asserts the S3 upload, datetime, or Pylon `mock_calls`. Previously the Notion-failure case verified all three. Add the same assertions back so a regression in side-effect ordering is caught.

**🟡 No request timeouts** — `hyperscribe/libraries/pylon.py:31, 66`
Both `requests_post` calls run without a `timeout`, so a hung Pylon endpoint can stall feedback submission indefinitely (synchronous handler in an ASGI worker). Add an explicit timeout (e.g. `timeout=10`).

**🟡 Trust boundary on `canvas-logged-in-user-id` header** — `capture_view.py:323`
The handler trusts a request header to look up the staff record and attach an email to the Pylon ticket. Confirm Canvas's reverse proxy strips/overwrites this header on inbound requests; if not, a user could spoof another staffer's email on tickets. Worth a one-line comment documenting the assumption either way.

**🟡 `dehyphenated` transform is lossy** — `pylon.py:18`
`"new-instance".replace("-", " ").title().replace(" ", "")` works for kebab-case but silently mangles already-mixed-case input: `"My-Instance"` → `"Myinstance"`. Probably fine for Canvas instance slugs (always lowercase kebab) but worth noting; consider `"".join(part.capitalize() for part in instance_name.split("-"))` for clarity.

**🟢 Minor: search loop logs every miss as `info` then logs the fallback** — When all three lookups miss, you get four `log.info` lines and no warning. Consider a single `log.warning` if the final fallback also returns `None`, since that's the only state the handler can't really recover from.

**🟢 Minor: `payload: dict` is untyped** — `pylon.py:60` uses bare `dict`. With `ty`/mypy strict this may flag; `dict[str, Any]` is more explicit.

alexiadowns-canvas:
Thanks @beaugunderson! 
I fed back all the items and they have been addressed in two commits (9e8896fb, db36b0de):

🔴 Inline import — moved to module-level imports
🟡 Stale failure-path assertions — restored S3, datetime, and Pylon mock assertions
🟡 Request timeouts — added timeout=10 to both requests_post calls
🟡 Trust boundary comment — added; header is set by Canvas's reverse proxy
🟡 Dehyphenated transform — switched to "".join(part.capitalize() ...) for clarity; also fixed the skip-dehyphenated test which was using an input that didn't actually skip
🟢 Search logging — consolidated to a single log.warning when all lookups (including fallback) miss
🟢 Untyped dict — now dict[str, Any]

Tested end-to-end after changes were made, feedback tickets are creating successfully in Pylon.

PR #257, opened 2026-05-04
May 3 Scribe enhancements: alert facility, PE/ROS reference + reconciliation, Charges matrix, robust UUID stamping

kevincareycanvas:
## Summary

Builds on `feat/canvas-scribe` with six threaded enhancement areas, all green through the CPA workflow:

### 1. Alert facility metadata on medication commands
- `medication_statement` and `stop_medication` commands now persist `alert_facility = "Yes" | "No"` via the two-phase originate-then-upsert-metadata pattern.
- Frontend medication cards expose a Canvas-navy checkbox (interactive when editable, static state-chip when read-only). Hidden on legacy commands where the field is undefined so existing notes don't see ghost UI.

### 2. PE/ROS — inline AI changes + previous-documentation reference
- Replaced per-subsection "updated from encounter" pills with a single info-circle trigger that reveals each section's prior encounter text inline.
- New "previous documentation" trigger surfaces the same provider's most recent finalized PE/ROS for this patient as a read-only reference, sourced from the stored CustomCommand HTML (base64-decoded) and parsed back into `{title, text}` sections via a sandbox-safe regex parser.
- Both triggers render in view AND edit modes; commits also strip `**asterisks**` from PE/ROS HTML on the way to Canvas. Same asterisk-strip applied to chart_review and history_review for consistency with the shared HistoryReviewRow editor.

### 3. Charges matrix with CPT↔ICD diagnosis pointers
- The flat 1-D Charges checklist is replaced with a 2-D matrix: diagnoses on rows, CPTs on columns, checkbox link cells (1–4 cap), drag-reorderable on both axes via two MIME types.
- Diagnosis rank is implicit in `commands[]` array position and propagates to `ClaimDiagnosisCode.rank` via Canvas's origination order — no backend write needed.
- CPT→ICD links are explicit, written to `BillingLineItem.assessment_ids` ("diagnosis pointers") via a new `ClaimLinkSync` handler that responds to `PERFORM_COMMAND__POST_COMMIT`. Translation chain: `linked_icd10_codes` → match against this note's `Assessment.condition.codings` → `Assessment.id`s → BLI assessment_ids.
- New `PerformBillingLineItemRemoval` handler responds to `PERFORM_COMMAND__POST_ENTER_IN_ERROR` and removes the matching BLI (filter on `cpt + note_id + status=ACTIVE` per SDK pattern) so the footer doesn't keep a stale row when a CPT is removed (× in matrix) or amended (void+recreate).
- Matrix is read-only during amendment to avoid mutating link state mid-flight; assess/diagnose deletion still routes through `pruneOrphanedLinks` so any perform that linked to the removed ICD gets re-edited (tag `_amend_edited`) and `ClaimLinkSync` re-fires on commit.
- Aesthetic: refined clinical-archive — hairline rules, tabular monospace, custom navy checkboxes, hover-revealed × remove, dot-grip drag indicators, single-line CPT description with ellipsis, centered headers, full-width call-to-action when no charges yet.

### 4. Robust UUID stamping on ScribeSummary post-insert
- Origin-tracked pipeline replaces the brittle display-string lookup. Each insertable item carries a `{src, srcIdx}` tag through the diagnose→assess conversion and RX-first sort, so post-insert UUID write-back lands on the exact source slot regardless of type conversions, ordering, or display collisions.
- Stamps `command_uuid` only — leaves cached `command_type` and `data` unchanged so the read-only Charges matrix still renders correctly post-approval.
- Fixes the cascading "Inserting N → All M inserted (N≠M)" mismatch that surfaced under multiple variants (recommendations, converted assesses, long-display commands).

### 5. Amendment workflow (KOALA-5485)
- "Make Changes" gesture on a finalized note transitions the Scribe tab into amend mode (`wasFinalized && !approved`). Editable sections defined by `EDITABLE_AMEND_SECTIONS` allowlist; clinical orders (`prescribe`, `refill`, `adjust_prescription`, `refer`, `imaging_order`, `lab_order`) are denied via `NON_EDITABLE_AMEND_COMMAND_TYPES` because their void+recreate has downstream external workflow consequences.
- Three amend routes selected per `section_key`: **DIRECT_EDIT** (chief_complaint — emit EDIT only, reuse uuid), **CUSTOM_COMMAND_ROUTED** (PE/ROS/chart_review/history_review/lab_results/imaging_results — EnterInError + Originate, home-app auto-commits CustomCommand), and **VOID_RECREATE** (everything else — EnterInError + Originate + Commit).
- Three frontend endpoints: `/edit-existing-commands` (POST tagged `_amend_edited` rows), `/delete-existing-commands` (POST tagged `_amend_deleted` rows, EIE only), and existing `/insert-commands` (genuinely new rows without command_uuid). Pipeline order: deletes first, then inserts, then edits — required so `ClaimLinkSync` on a recreated perform can resolve `linked_icd10_codes` against newly-Originated Assessments before the perform's own commit fires.
- Mutual exclusion between `_amend_edited` and `_amend_deleted` enforced at source (`handleEdit`, `pruneOrphanedLinks`) and at sink (`amendEdits` filter) to prevent doubly-tagged rows from being both EIE'd by /delete and re-Originated by /edit.
- × on an on-note diagnose during amendment now routes through `onDeleteCommand` (tag `_amend_deleted`) instead of `onEditCommand` (which previously stamped `data.rejected=true` and `_amend_edited` — `DiagnoseParser.build` ignores `data.rejected` so the same ICD was being recreated under a new uuid).

### 6. Hardening
- `/verify-commands` now requires note-author authorization (matching `/insert-commands` and `/insert-metadata`); also tolerant of malformed `attempted` payloads. UUID-syntax validation via `_is_uuid` guard prevents Django UUIDField coercion 500s.
- ClaimLinkSync uses `prefetch_related("condition__codings")` to collapse a per-assessment N+1 to 2 queries; assessment query excludes `entered_in_error` rows so amend re-fires see a clean state.
- `prior_sections.py` Note read trimmed via `.only("id", "patient_id", "provider_id")` and verbose diagnostic logging removed.
- Dead `onAddCharge` prop chain (replaced by inline `AddChargeCell`) removed across `summary.js` + `soap-group.js`.
- Defensive: pre-PR signed notes (command_uuid without `already_documented` flag) handled correctly via the `(already_documented || command_uuid)` OR convention used in `isInsertableCommand`, `handleMakeChanges`, `handleDelete`, `rowLocked`, and `unlinkedCptCount`.

## Verification

- [x] `mypy --config-file=mypy.ini .` — 0 errors
- [x] `pytest tests/` — 2036+ tests pass, ≥90% coverage
- [x] New code coverage: ClaimLinkSync 100%, PerformBillingLineItemRemoval ≥90%, prior_sections 99%, _section_html 91%
- [x] Security review: auth on `/verify-commands`, UUID syntax guards, payload robustness
- [x] DB performance review: N+1 prefetch, narrowed Note read, entered_in_error filter on assessment query
- [ ] Manual UAT on scribeqa-playground covering: alert facility toggle, PE/ROS prior-visit reference, Charges matrix happy path (link → sign → claim shows pointers), reorder rank, reject diagnosis (auto-prune), Accept & Sign with unlinked CPT (blocked), Make Changes amendment (edit/delete across each section bucket, charges matrix read-only, no spurious re-Originate of deleted diagnoses)

## Reports

CPA workflow artifacts in `.cpa-workflow-artifacts/` (local-only, gitignored):
- `coverage-report-*.md`
- `security-review-*.md`
- `db-performance-review-*.md`
- `wrap-up-report-*.md`


kevincareycanvas:
@claude review

kevincareycanvas:
Both bot-flagged bugs addressed in 37cdfc95.

**1. HTML entity decoding in `_section_html.py`** — Added a sandbox-safe `_decode_html_entities` helper that handles both numeric (`&#NNN;` / `&#xHH;`) and named (`&amp;`, `&lt;`, `&gt;`, `&quot;`, `&apos;`) refs. Numeric refs are decoded first so any literal `&` they produce isn't reinterpreted in the named-ref pass. Applied to both title and text in `parse_ros_pe_html`. 6 new tests cover decimal, hex, named refs, entities-in-title, extended Unicode (µ/em-dash/é), and the malformed-codepoint passthrough.

**2. `buildRankedDiagnoses` includes assess commands** — Loosened the filter to accept both `diagnose` and `assess` types when `data.icd10_code` is populated. Diagnoses still require `accepted` (assess is implicitly accepted — it's an action being taken). Assess commands created by "+ Add Condition" with a `conditionId` now appear as rows in the matrix and the chronic-condition follow-up workflow no longer hits the permanently-disabled Accept & Sign trap. Backend ClaimLinkSync already resolves either type via `Assessment.condition.codings`, so this is UI-only.

Net: +82 / −5 across `_section_html.py`, `summary.js`, and `test_section_html.py`. All 2042 tests pass; coverage holds.

kevincareycanvas:
Round 2 review feedback addressed in 2292f06d.

**1. ClaimLinkSync ordering race** (🔴) — `build_effects` runs all originates then all commits, so `commands=[perform, assess]` would emit `[orig_perform, orig_assess, commit_perform, commit_assess]` and ClaimLinkSync would fire on `commit_perform` before the Assessment exists in the DB. The reachable workflow you flagged (empty A&P → user adds CPT first → "+ Add Condition" appends assess at end) survives the previous RX-first sort because non-RX commands stayed in input order. **Fix:** extended the sort in `handleInsert` from a single RX-first class to four priority classes:

| Class | Types |
|---|---|
| 0 | `prescribe` / `refill` / `adjust_prescription` (RX top-of-note, unchanged) |
| 1 | `diagnose` / `assess` (must precede perform) |
| 2 | everything else |
| 3 | `perform` (must follow A&P) |

`Array.prototype.sort` is stable in modern JS, so within a class the user's original order is preserved. ClaimLinkSync's docstring updated to document this is now an actively-enforced invariant.

**2. Orphan prune too narrow** (🔴) — replaced `icdToPrune` (which only caught accepted→rejected toggles) with a general `pruneOrphanedLinks(prev, next)` helper that computes the set diff of `buildRankedDiagnoses` on prev vs. next and strips any orphaned ICD from every `perform.linked_icd10_codes`. Helper is called from both `handleEdit` and `handleDelete`, so all three flagged paths — ICD change via search dropdown (clear+select two-step), ICD clear in isolation, and assess/diagnose × deletion — are covered uniformly. Robust to any future code path that mutates the ranked-diagnosis set.

**3. Audit-log denial on /verify-commands** (🟡) — added `audit_event(note_uuid, "VERIFY_COMMANDS_DENIED", {})` before the 403 return, parallel to `/insert-metadata`'s `INSERT_METADATA_DENIED`. Test `test_verify_commands_denies_non_author` extended to assert the audit_event fires.

Net: +84/−36 across `session_view.py`, `claim_link_sync.py` (docstring only), `summary.js`, and `test_session_view.py`. All 2042 tests pass; mypy clean.

kevincareycanvas:
Round 3 review feedback addressed in b08bd9b4. Four findings, four fixes:

**1. `/verify-commands` editability regression** (🔴) — My round-2 `_authorize_edit` addition inherited the `editable_note` check, which 403s post-sign reloads by the actual author. The auto-verify-on-load effect calls verify-commands on every reload of an approved note, so this broke the legit workflow: `fetch` resolves on 4xx, `data.failed?.length || 0` evaluates to 0, and the banner rendered a misleading **"All 0 command(s) inserted successfully"** plus a `VERIFY_COMMANDS_DENIED` audit row per legit reload. **Fix:** extracted a narrower `_authorize_read_as_author` helper (staff-vs-provider check only, no editability gate) and use it here. Read-only endpoint shouldn't be gated on editability. New test `test_verify_commands_allows_author_after_note_signed` asserts the legit author still gets through with `editable_note → False`.

**2. `MedicationRow` form Save silently reverts the view-mode `alert_facility` toggle** (🔴) — `useState(!!command.data.alert_facility)` only honors its mount value; toggling the new view-mode checkbox updated `command.data` but local form state stayed stale. Open the form to edit, click Save → the stale local overwrites the up-to-date data. **Fix:** one-line `useEffect(() => setAlertFacility(!!command.data.alert_facility), [command.data.alert_facility])` near the state declaration. Mirrors how `handleCancel` already syncs from props.

**3. `_wait_for_command` lacks `try/except` around `.exists()`** (🟡) — A transient `OperationalError` would escape the helper, escape `build_metadata_effects` (whose per-item `try/except` only wraps `build_stub` + `upsert_metadata`, not the wait that precedes it), 500 `/insert-metadata`, and the frontend's `catch (metaErr)` swallows it silently. Every alert_facility metadata write in the batch silently lost. **Fix:** wrap the `.exists()` in `try/except`, treat failures as "not yet visible" so the existing per-item skip path takes over via the documented degradation.

**4. `prior_sections` test mocks on the wrong ORM chain** (🟡) — When I switched production to `Note.objects.only(...).get(id=...)` in the wrap-up cleanup, I didn't update the test mocks. They still set `mock.objects.select_related.return_value.get...` or `mock.objects.get`. MagicMock auto-creates fresh chains for unconfigured paths, so the seeded fixtures were never consulted; tests passed by coincidence. **Fix:** all setups now use `mock.objects.only.return_value.get`, so the tests actually exercise the seeded values.

| Metric | Before | After |
|---|---|---|
| Tests | 2042 | **2043** (+1 new for post-sign happy path) |
| mypy | 0 errors | 0 errors |
| Diff | — | +114 / −23 across 5 files |

kevincareycanvas:
Round 4 review feedback addressed in 59e4a101.

**🔴 `.exclude(note_id=note.id)` UUID-vs-bigint mismatch** — confirmed by empirical reproduction:

```
uuid: bd8316a0-6791-42a7-80ca-d7f6756a30b5
int(uuid): 251904741013387010038717631669419520181   (128-bit)
sql compiles OK; eval FAILS: OverflowError (SQLite test) / bigint range error (Postgres prod)
```

`Command.note` is a default Django FK targeting `Note.dbid` (BigAutoField PK from `IdentifiableModel`'s base `Model`). `Note.id` is a unique-constrained UUIDField, **not** the PK. Passing `note.id` (a `uuid.UUID` object) into `.exclude(note_id=...)` flows through `BigIntegerField.get_prep_value` → `int(uuid)` → 128-bit integer parameter → DB rejects on bigint range. Exception escapes `.first()`, caught by the surrounding `try/except Exception`, logged as `prior_sections: query failed for note <uuid>`, and the function returns the empty payload.

Visually indistinguishable from "no prior PE/ROS for this patient" — which is why the UAT testing didn't catch it. The codebase convention is unambiguous and this PR follows it correctly everywhere else (`claim_link_sync.py` uses `note_id=note.dbid` for ScribeSummary, Assessment, BillingLineItem); this single line was the deviation.

**Fix** (one character): `note.id` → `note.dbid`. Also extended the `.only(...)` hint to include `"dbid"` so accessing it doesn't trigger an extra column SELECT.

**Regression guard** added in `test_returns_payloads_for_each_command_type`:

```python
mock_command_cls.objects.filter.return_value.exclude.assert_called_once_with(note_id=778)
```

All fixture notes gain `dbid=778` to support that assertion. Without this guard, MagicMock's permissive kwarg behavior would let the same class of regression through again.

| Metric | Result |
|---|---|
| Tests | **2043 pass** (no count change; existing prior_sections happy-path test now does what it claimed to do) |
| mypy | 0 errors |
| Diff | +57 / −11 across 2 files |

Reply: https://github.com/canvas-medical/canvas-hyperscribe/pull/257#issuecomment-...

Empirical verification log appended to local `.cpa-workflow-artifacts/`.

kevincareycanvas:
Round 5 review feedback addressed in b66e2d45.

**🔴 ClaimLinkSync diagnosis-pointer order scrambled by `set`** — the bot's analysis is correct and I had previously observed the symptom but mis-categorized it. The handler used `wanted: set[str]` and iterated it to build `assessment_ids`. Python set iteration is hash-randomized, so `BillingLineItem.assessment_ids` ordering — which **is** the diagnosis-pointer sequence on CMS-1500 box 24E (primary, secondary, …) — was effectively random.

Single-ICD CPTs were unaffected (1-element list = 1-element set), so the bug was invisible in quick-test paths. Multi-ICD CPTs (E&M visit + chronic-condition follow-up — exactly the workflow this PR's matrix targets) silently flipped primary-vs-secondary intent. The frontend's `handleToggleCptLink` explicitly preserves provider click order with `next = [...current, icdCode]`; my backend was destroying that signal at the very first aggregation point.

My tests previously had `sorted(call_kwargs["assessment_ids"]) == ["a-1", "a-2"]` and a comment `# Two distinct assessments, in any order.` — I observed the nondeterminism and accepted it instead of recognizing it as a real ordering invariant. The bot's comment caught exactly that.

**Fix** — replace the set with an order-preserving deduped list:

```python
wanted: list[str] = []
seen: set[str] = set()
for entry in summary.commands:
    ...
    for c in links:
        stripped = _strip(c)
        if stripped and stripped not in seen:
            seen.add(stripped)
            wanted.append(stripped)
```

The downstream `target_assessment_ids = [icd_to_assessment[icd] for icd in wanted ...]` now preserves the click order all the way through to the BLI write.

**Test tightening** — both `sorted(...)` assertions changed to exact list equality:
- `test_compute_emits_update_with_translated_assessment_ids` now asserts `["a-1", "a-2"]` (matching `linked_icd10_codes = ["E11.9", "I10"]`).
- `test_compute_dedupes_links_across_multiple_perform_entries_for_same_cpt` now asserts `["a-1", "a-2"]` (the merge of `[E11.9]` and `[E11.9, I10]` in append-if-not-seen order).

**New regression guard** — `test_compute_preserves_provider_click_order_for_diagnosis_pointers` feeds four codes in reverse-alphabetical order (`["M54.5", "K21.9", "I10", "E11.9"]`) with assessments returned in a different order, and asserts the click order survives all the way to `assessment_ids`. A `set` or `sorted` impl would fail this test.

| Metric | Result |
|---|---|
| Tests | **2044 pass** (+1 new regression guard) |
| mypy | 0 errors |
| Diff | +84 / −8 across 2 files |

kevincareycanvas:
@claude review

kevincareycanvas:
@claude review once

kevincareycanvas:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #263, opened 2026-05-08
Fix generate-summary mode preservation by falling back to the DB

alexiadowns-canvas:
Completes the fix from #261 — the generate-summary endpoint now falls back to                                                        existing DB values for `mode` and `selected_template_name` when the caller doesn't include them in the request body.


alexiadowns-canvas:
@claude review

PR #269, opened 2026-05-12
Attach alert_facility metadata before commit (gated by AlertFacilityEnabled)

kevincareycanvas:
## Summary

Minimal surgical change off `feat/canvas-scribe` to store `alert_facility` metadata on `medication_statement` and `stop_medication` commands, attached to the command **before** the commit effect transitions it from STAGED → COMMITTED.

- **Format**: `alert_facility="Yes"` (when the UI toggle is on) or `"No"` (default when the feature flag is on but the toggle is off). Matches the `alert_facility_fields` plugin contract.
- **Server-side gate**: backend reads the `AlertFacilityEnabled` plugin secret. When unset → **zero** metadata effects emitted. When set → emit `"Yes"`/`"No"` on every medication_statement / stop_medication.
- **Timing**: metadata effect is inserted into the returned effect list between originate and commit. Canvas's sequential effect processor applies originate first (row appears as STAGED), then the metadata (apply-time SDK check passes), then commit (transitions to COMMITTED with metadata attached). All in a single round-trip.
- **SDK bypass**: `command.upsert_metadata(...)` runs a synchronous `Command.objects.filter(...).exists()` check at python-build time, which always fails before originate has been applied. The fix constructs `UPSERT_COMMAND_METADATA` as a raw protobuf Effect directly. A test pins the exact payload shape (effect type + JSON keys) so any future SDK schema change for that effect breaks CI rather than silently diverging in production.

## What this diff does NOT touch

- **Zero frontend changes**: no JS / CSS / static assets modified. The frontend already sends `data.alert_facility: true/false` and gates UI on the flag — both work unchanged.
- **No new constants** in `hyperscribe/libraries/constants.py`. The literals appear in 2 parsers and 1 test pin; centralising them would add diff without type safety.
- **No removal of `build_metadata_effects` / `/insert-metadata`**: kept byte-for-byte from baseline. `build_effects` now always returns an empty `metadata_pending`, so the legacy two-phase path becomes a no-op for alert_facility but is still available for any other future use.
- **No new audit events** in `session_view.py`. The existing `INSERT_COMMANDS` event already captures what matters.
- **No manifest change**. `AlertFacilityEnabled` plugin secret already exists in baseline.

## Files modified

| File | Purpose |
|---|---|
| `hyperscribe/scribe/commands/base.py` | Add `feature_flags` kwarg to default `pending_metadata` signature |
| `hyperscribe/scribe/commands/medication_statement.py` | Update `pending_metadata` to gate on flag and emit Yes/No |
| `hyperscribe/scribe/commands/stop_medication.py` | Same |
| `hyperscribe/scribe/commands/builder.py` | Add `_build_unvalidated_metadata_effect` helper; emit metadata inline between originate and commit |
| `hyperscribe/scribe/api/session_view.py` | One line: build `feature_flags` from `self.secrets` and pass to `build_effects` |
| Tests | Pinning the payload shape + flag on/off + Yes/No defaults for both command types |

`git diff --stat origin/feat/canvas-scribe..HEAD`: 9 files changed, 242 insertions(+), 46 deletions(-)

## Test plan

- [x] `uv run pytest tests/` — all 2003 tests pass
- [x] `uv run mypy hyperscribe/` — clean (198 files)
- [x] Manual sanity in test environment:
  - With `AlertFacilityEnabled` unset → insert a medication command, confirm no `alert_facility` row exists in the Canvas `command_metadata` table for that command.
  - With `AlertFacilityEnabled` set → insert two medication commands (one toggle on, one off). Both have an `alert_facility` row, values `"Yes"` and `"No"` respectively. Same for stop_medication.
  - Confirm metadata is attached on the same response as the commit (no second-phase round-trip; visible immediately on first read).
  - Confirm no other command type writes an `alert_facility` row.

## Risk

The raw protobuf bypass deliberately skips one SDK build-time validator. The dedicated test pins the exact payload shape (effect type literal + JSON schema), so a future SDK breaking change for `UPSERT_COMMAND_METADATA` produces a clean CI failure rather than silent divergence. No native SDK alternative exists for attaching metadata before commit within a single effect-handler response.

### DIFF
--- a/hyperscribe/scribe/api/session_view.py
+++ b/hyperscribe/scribe/api/session_view.py
@@ -1115,7 +1115,8 @@ class ScribeSessionView(StaffSessionAuthMixin, SimpleAPI):
                     status_code=HTTPStatus.BAD_REQUEST,
                 )
             ]
-        effects, metadata_pending, attempted = build_effects(commands, note_uuid)
+        feature_flags = {"AlertFacilityEnabled": bool(self.secrets.get("AlertFacilityEnabled"))}
+        effects, metadata_pending, attempted = build_effects(commands, note_uuid, feature_flags)
         audit_event(
             note_uuid,
             "INSERT_COMMANDS",
--- a/hyperscribe/scribe/commands/base.py
+++ b/hyperscribe/scribe/commands/base.py
@@ -79,8 +79,13 @@ class CommandParser(ABC):
         except Exception:
             return []
 
-    def pending_metadata(self, command: _BaseCommand, proposal: dict[str, Any] | None = None) -> dict[str, Any] | None:
-        """Return metadata to upsert in phase 2, or None. Override per command type."""
+    def pending_metadata(
+        self,
+        command: _BaseCommand,
+        proposal: dict[str, Any] | None = None,
+        feature_flags: dict[str, bool] | None = None,
+    ) -> dict[str, Any] | None:
+        """Return metadata to upsert, or None. Override per command type."""
         return None
 
     def build_stub(self, command_uuid: str, note_uuid: str) -> _BaseCommand:
--- a/hyperscribe/scribe/commands/builder.py
+++ b/hyperscribe/scribe/commands/builder.py
@@ -4,7 +4,6 @@ import json
 import uuid
 from typing import Any
 
-from canvas_generated.messages.effects_pb2 import Effect as _RawEffect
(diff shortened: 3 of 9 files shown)


kevincareycanvas:
@claude review

kevincareycanvas:
Successfully tested and passed claude code review. Merging.

PR #274, opened 2026-05-15
fix: stop the Scribe save-summary 403 storm from _authorize_edit

nmpsilva:
[KOALA-5475](https://canvasmedical.atlassian.net/browse/KOALA-5475)

## Summary

Background `save-summary` was firing on every Scribe state change. Once the user's Django session expired (typical on iPad with idle disconnects), each subsequent change fired another POST that 403'd at home-app's middleware — ~2,727 per week on Brigade, ~1.9% of save-summary traffic. The frontend's existing try/catch only catches network errors, not HTTP failures, so each 403 was silently ignored client-side while still hitting the server and the provider had no idea their edits weren't persisting.

This PR adds:

1. **A gate (sessionLostRef + state).** On the first 401/403, pause further autosaves this focus cycle. A visibilitychange listener clears the gate on focus return so a user who re-authenticated elsewhere resumes autosaving without a forced reload. If the session is still bad, the next save trips the gate again — at most one 403 per focus cycle instead of one per edit.
2. **A red "session expired" banner** at the top of the Scribe tab, surfaced as soon as the gate trips. Provider sees the failure mode immediately instead of silently accumulating edits that won't save. Uses a new `.readonly-banner--alert` style so it reads more urgently than the existing neutral locked/non-author banners.
3. **A `!isAuthor` early-return** in `saveSummaryToCache`. Non-author viewers legitimately 403 on `_authorize_edit`; suppressing the request entirely avoids a confusing "session expired" alarm for what is really an authorization mismatch (their existing read-only banner already covers this case).

## Out of scope (worth filing as follow-ups)

- **Server-side graceful auth.** The fundamental fix would be to drop StaffSessionAuthMixin for save-summary and authenticate at the plugin-runner level. `ScribeSummary` is a `CustomModel` in the plugin's own namespace, so the data layer doesn't need the user session; the session gate sits only at the HTTP route. Implementing that means designing a new trust model (signed note-token + plugin-runner credentials), which is bigger than this ticket and needs sec/product input. Filed as a follow-up.
- **localStorage cache for offline recovery.** The current banner makes the failure visible so the provider can reload and re-auth — they still lose edits made between session expiry and reload. A localStorage replay layer would close that gap. Separate change with HIPAA considerations.
- **The user-visible "Response not successful: 403" popup** referenced in the ticket comes from the parent home-app's Apollo client polling, not from this fetch. That's a host-app concern, not in this plugin.

## Files changed

**Logic (3 files):**
- `hyperscribe/scribe/static/summary.js` — gate, banner state, visibilitychange handler, `!isAuthor` short-circuit.
- `hyperscribe/scribe/static/styles.css` — `.readonly-banner--alert` red/amber variant.

**Incidental:** 11 unrelated scribe modules reformatted by `ruff format` (the prior commit on this branch skipped them; including them here so future rebases stay clean).

## Test plan

- [x] `uv run ruff format .` / `uv run ruff check .` / `uv run mypy .` / `uv run pytest tests/` — clean.
- [x] Manual UAT against local Canvas via Playwright + DevTools:
  - **Author + forced 403:** red banner appears with exact copy, console warns once, only one save-summary fires.
  - **Subsequent edits with gate set:** no further save-summary requests.
  - **visibilitychange + real fetch + state change:** save fires (200), banner clears.
  - **Non-author iframe (Beau's note viewed as Larry):** only the regular read-only banner; no false alert; zero save-summary requests originated.

Tested in a fresh patient record with both author and non-author note iframes loaded simultaneously to confirm the banners stay scoped correctly.

[KOALA-5475]: https://canvasmedical.atlassian.net/browse/KOALA-5475?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/hyperscribe/scribe/api/session_view.py
+++ b/hyperscribe/scribe/api/session_view.py
@@ -215,9 +215,7 @@ def _save_summary(note_id: str, payload: dict[str, Any]) -> None:
 
 def _infer_mode_for_heal(note_dbid: int, has_user_content: bool) -> str:
     """Pick the user's mode from session artifacts. '' means no signal — don't heal."""
-    events = (
-        ScribeAuditLog.objects.filter(note_id=note_dbid).values_list("events", flat=True).first()
-    ) or []
+    events = (ScribeAuditLog.objects.filter(note_id=note_dbid).values_list("events", flat=True).first()) or []
     last_start = ""
     for event in events:
         event_type = event.get("type") if isinstance(event, dict) else None
@@ -245,10 +243,7 @@ def _has_user_authored_content(row: dict[str, Any]) -> bool:
     if row["note_data"]:
         return True
     commands = row["commands"] or []
-    return any(
-        isinstance(cmd, dict) and not cmd.get("_template_inserted")
-        for cmd in commands
-    )
+    return any(isinstance(cmd, dict) and not cmd.get("_template_inserted") for cmd in commands)
 
 
 def _load_summary(note_id: str) -> dict[str, Any] | None:
--- a/hyperscribe/scribe/application/transcript_app.py
+++ b/hyperscribe/scribe/application/transcript_app.py
@@ -62,11 +62,7 @@ def _scribe_tab_has_content(note_dbid: int) -> bool:
     if items:
         return True
 
-    summary = (
-        ScribeSummary.objects.filter(note_id=note_dbid)
-        .values("note_data", "commands", "approved")
-        .first()
-    )
+    summary = ScribeSummary.objects.filter(note_id=note_dbid).values("note_data", "commands", "approved").first()
     if not summary:
         return False
     if summary["approved"] or summary["commands"]:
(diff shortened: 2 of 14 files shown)


kevincareycanvas (changes requested):
@nmpsilva the UX needs to be more explicit to inform/guide the user.

The banner sits at the very top of the page and easily and is out of view for the majority of the scribe tab view. A provider editing Plan/Assessment at the bottom of a long note won't see it, can keep typing, hit save, and walk away thinking the note is documented when nothing persisted.

Asks:

1. The alert must be visible from anywhere in the note. I don't have a strong opinion on the form — sticky banner, floating toast, modal overlay, anchored side indicator are all fine. The requirement is that a provider can't be on any part of the Scribe surface without seeing it.

2. Make the Scribe surface read-only on 401 — same treatment as locked/non-author notes. Today the autosave gate stops POSTs but inputs stay enabled, so a provider can still type into a dead form.

3. Update the copy to "Session expired. Edits aren't being saved. Refresh the page to proceed." Concise and the action is unambiguous.

Out of scope for this PR but worth filing: local-storage draft preservation and a keep-alive ping to avoid the expiry firing at all.

PR #287, opened 2026-06-02
fix: dedupe approved Scribe commands from Additional Commands (KOALA-5634)

nmpsilva:
[KOALA-5634](https://canvasmedical.atlassian.net/browse/KOALA-5634)

On Scribe summary approval, commands that successfully posted to the Commands tab were appearing duplicated under the **Additional Commands** section of the Summary tab — most commonly `medication_statement` recommendations and `diagnose` cards whose ICD matched a patient condition. The duplicates also weren't rendering as already-documented. Brigade-priority defect.

Three root causes, all in `summary.js`. (1) Accepted recommendations never had their server uuid stamped locally on `/insert-commands` success — only `workingCommands` did — so the next `syncNoteCommands` sweep saw "new" uuids and dropped them into the `from_the_note` fallback. (2) The diagnose→assess transformation flipped `command_type` on the POST but not on the local row, so the canonical `${command_type}:${display}` lookup never matched and stamping silently missed. (3) `syncNoteCommands` only consulted `commands`, not `recommendations`, so even a correctly-stamped rec would re-surface.

Fix extracts a `commitRecommendations(next)` helper that synchronously primes a `recommendationsRef` before `setRecommendations` at the three approval-flow sites (cache-load, approve, `handleAddNow`) — closing the React-commit race against `syncNoteCommands`. Diagnose rows that match a patient condition are now flipped to `command_type: 'assess'` upstream of the stamping pass so the canonical lookup key matches without alias bookkeeping (and without the 80-char-truncated `display` collision surface the alias would have introduced). Four structural-pin tests in `test_session_view.py` pin the invariants. Also fixes a pre-existing fixture bug in `test_edit_existing_commands_silently_drops_disallowed_section` whose prescribe payload lacked required `substitutions`/`fdb_code` fields, causing `validate_proposals` to 400 before the section-key filter could exercise the silent-drop contract under test.

Localhost UAT passed: approved a Scribe summary containing both a `medication_statement` recommendation and a `diagnose` with an ICD matching the patient's active condition; post-approval `from_the_note` contained only pre-existing ad-hoc chart commands (no duplicates of the just-approved entries), `already_documented=True` on the stamped rec.

[KOALA-5634]: https://canvasmedical.atlassian.net/browse/KOALA-5634?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/hyperscribe/scribe/commands/_rx_validation.py
+++ b/hyperscribe/scribe/commands/_rx_validation.py
@@ -25,9 +25,11 @@ front-end, and makes the failure debuggable from the audit log.
 This module is intentionally framework-agnostic: pure functions that take a
 ``data`` dict and return a list of human-readable error strings.
 """
+
 from __future__ import annotations
 
 import re
+
 # NOTE: only ``Decimal`` is imported here. The Canvas plugin sandbox allowlists
 # ``Decimal`` from ``decimal`` but rejects ``InvalidOperation`` — importing it
 # raises ``ImportError`` at plugin-load time and de-registers every handler
@@ -58,9 +60,7 @@ def _has_value(value: Any) -> bool:
     return True
 
 
-def _validate_required_string(
-    data: dict[str, Any], field: str, label: str, errors: list[str]
-) -> None:
+def _validate_required_string(data: dict[str, Any], field: str, label: str, errors: list[str]) -> None:
     if not _has_value(data.get(field)):
         errors.append(f"{label} is required")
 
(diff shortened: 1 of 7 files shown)


kevincareycanvas:
UATing this now

kevincareycanvas:
@nmpsilva this appears to have done the trick for the classic scribe experience where the note is only signed once. 

During UAT I found a couple of issues that that are outside the scope of this PR. They are mainly related to the amendments workflow. I will document these in jira and have them prioritized for you.

PR #288, opened 2026-06-02
feat: surface background carry-forward on recommended diagnoses (KOALA-5635)

nmpsilva:
[KOALA-5635](https://canvasmedical.atlassian.net/browse/KOALA-5635)

Surfaces the background carry-forward field on **Recommended Diagnosis** cards in the Scribe Summary tab. Previously this field only appeared on manually-added Assess Condition cards; Nabla-suggested A&P entries (rec-diagnose) had no `condition_id` stamped at split time and the prefill function filtered to `command_type='assess'` only, so the textarea never had data to bind to.

Backend: `split_plan_into_diagnoses` now looks up active patient conditions and stamps `data.condition_id` on diagnose proposals whose ICD-10 matches (normalized: dots stripped, uppercased — mirrors the frontend `handleInsert` match). A parallel `prefill_diagnose_backgrounds` in `builder.py` delegates to the existing `carry_forward_assess_background` helper. Frontend `DiagnoseRow` renders an editable Background textarea above today's-assessment, gated on `data.condition_id` being truthy.

On ICD change (`handleSelect`) or clear (`handleClearCode`), both `condition_id` and `background` are explicitly cleared after the spread. This prevents a cross-attachment bug where preserved background text scoped to the OLD condition could land against a NEW condition's Assessment via `handleInsert`'s diagnose→assess flip. The user-accepted tradeoff: typed text is lost on any ICD change (including within-condition typo fixes) in exchange for a clinical-data-integrity guarantee.

Sandbox-aware exception narrowing: `django.core.exceptions.ValidationError` is not in `plugin_runner/sandbox.py` ALLOWED_MODULES, so an upfront `uuid.UUID(str(note_uuid))` pre-validation is the sandbox-safe equivalent of a broad except for malformed-uuid handling.

Localhost UAT verified: rec-diagnose card renders the textarea with prefilled content when condition matches; clicking the ICD prefix clears the field; selecting a non-matching new ICD doesn't surface the old background. Pre-existing fixture fix from #287 bundled to unblock CI until that PR merges. Known followups: dormant `ValidationError` gap in sibling `prefill_assess_backgrounds*` functions; ICD-9 vs ICD-10 filter tightening.

[KOALA-5635]: https://canvasmedical.atlassian.net/browse/KOALA-5635?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/hyperscribe/scribe/api/session_view.py
+++ b/hyperscribe/scribe/api/session_view.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import json
 import re
+import uuid
 from datetime import datetime, timezone
 from http import HTTPStatus
 from typing import Any, Union
@@ -64,6 +65,7 @@ from hyperscribe.scribe.commands.builder import (
     build_metadata_effects,
     prefill_assess_backgrounds,
     prefill_assess_backgrounds_for_proposals,
+    prefill_diagnose_backgrounds,
     validate_proposals,
 )
 from hyperscribe.scribe.commands.carry_forward import carry_forward_assess_background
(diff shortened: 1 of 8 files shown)


kevincareycanvas:
I am UATing this now.

kevincareycanvas:
@nmpsilva this is looking good so far! 

Before we merge, we need to make changes to keep the documentation experience consistent. 

The main change is that the background field needs to be available in **ALL** condition cards for **ALL** of the following situations.

- New diagnosis (diagnose command) 
- Assessing a preexisting condition (assess command)
- The card is an AI recommendation
- The card is user initiated entry
- The most recent background field is blank for this condition
- The most recent background field is **not** blank for this condition 
- When an ICD10 code is changed in the condition card

Main take away: The background field must always be available for the end user to leverage when the scribe summary tab is being edited.

Also lets make some subtle tweaks to the UI to indicate that background is optional, user maintained, and carries forward. Here is a mocked up v2

image

Background field should be shorter than today assessment. Also add help text under background to assist users in navigating the change.

kevincareycanvas:
This looks good and tested well. Going to merge. Did detect an unrelated issue are command insertion into the note body during amendment flow but this is outside the scope of this PR. Will create a new Jira ticket. Future enhancements to this feature will involve improving the protection around background content during diagnosis code alterations.

PR #289, opened 2026-06-03
fix: stop gating /note-commands on note editability (KOALA-5689)

nmpsilva:
[KOALA-5689](https://canvasmedical.atlassian.net/browse/KOALA-5689)

After approving the Scribe tab, adding commands on the Commands tab, and signing the note, the Scribe tab's **ADDITIONAL COMMANDS** section didn't reflect the newly added commands until the user clicked Amend. Root cause: `syncNoteCommands` (`summary.js:629`) calls GET `/note-commands` on every NOTE_TAB_CHANGE, but `/note-commands` was gated by `_authorize_edit` which returns 403 once `Helper.editable_note()` is False (post-sign). The client bails on any non-200 (`if (!res.ok) return;`), so the sync silently no-oped post-sign — stale ADDITIONAL COMMANDS until amendment re-opened the edit gate.

Fix: add `_authorize_read` mirroring `_authorize_edit` minus the `editable_note()` check, and switch `/note-commands` to use it. Keeps the note-exists check and the staff-is-provider check intact — only the editability gate is dropped, because `/note-commands` is a read-only display sync that has no business requiring the note to be editable. No other endpoint switches helpers; scope is intentionally narrow.

Tests: 7 unit tests on `_authorize_read` (4xx error paths + the signed-note-with-correct-staff bug scenario), 1 structural pin asserting `get_note_commands` uses `_authorize_read` and not `_authorize_edit`, 1 DB+factories integration test driving a real `NoteStateChangeEvent` to SIGNED and asserting `get_note_commands` returns 200 with the chart-side commands.

[KOALA-5689]: https://canvasmedical.atlassian.net/browse/KOALA-5689?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/hyperscribe/scribe/api/session_view.py
+++ b/hyperscribe/scribe/api/session_view.py
@@ -133,6 +133,38 @@ def _authorize_edit(note_uuid: str, request: Any) -> JSONResponse | None:
     return None
 
 
+def _authorize_read(note_uuid: str, request: Any) -> JSONResponse | None:
+    """Return a JSONResponse to short-circuit when the staff cannot read the note's Scribe tab.
+
+    Mirrors ``_authorize_edit`` but drops the ``Helper.editable_note()`` gate —
+    read endpoints (e.g. ``/note-commands``) need to keep working after the
+    note is signed, so the Scribe tab can reflect ad-hoc commands added on the
+    Commands tab post-sign without forcing the user to amend (KOALA-5689).
+
+    Authorized when the note exists and the logged-in staff (from
+    ``canvas-logged-in-user-id`` header) matches the note's provider. Returns
+    ``None`` on success.
+
+    Uses ``.values("provider__id")`` — no ``dbid`` needed since we skip the
+    editable_note() check.
+    """
+    if not note_uuid:
+        return JSONResponse({"error": "note_uuid is required"}, status_code=HTTPStatus.BAD_REQUEST)
+    try:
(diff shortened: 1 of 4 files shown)


hyperscribe/scribe/api/session_view.py
```diff
@@ -1993,7 +2025,11 @@ def get_note_commands(self) -> list[Union[Response, Effect]]:
         note_uuid = str(self.request.query_params.get("note_id", ""))
         if not note_uuid:
             return [JSONResponse({"error": "note_id is required"}, status_code=HTTPStatus.BAD_REQUEST)]
-        if denial := _authorize_edit(note_uuid, self.request):
+        # KOALA_5689_NOTE_COMMANDS_READ_AUTH: read-only display sync — must keep
+        # working post-sign so ad-hoc commands added on the Commands tab surface in
+        # the Scribe tab without forcing the user to amend. Uses _authorize_read
+        # (no editable_note gate) instead of _authorize_edit.
+        if denial := _authorize_read(note_uuid, self.request):
             return [denial]
```
ruipbatista, line 2042: this if statement reads like:

```python
if _authorize_read:
   ACCESS_DENIED
```

kevincareycanvas:
Successfully UATed with no issues. Going to merge now.

PR #297, opened 2026-06-11
feat: block note lock until Scribe tab is finalized (KOALA-5622)

kevincareycanvas:
## Summary

Prevents a provider from signing/locking a note that was documented with Scribe while the Scribe tab is still **un-finalized** — either before the AI summary is approved, or after an amendment has been started but not re-approved. Notes that never used Scribe are unaffected.

## How it works

- New handler `NoteLockGuard` (`hyperscribe/scribe/handlers/note_lock_guard.py`) responds to `NOTE_STATE_CHANGE_EVENT_PRE_CREATE` and returns an `EventValidationError` to abort the lock (message surfaced in the UI).
- "Finalized" maps to **`ScribeSummary.approved`** — the frontend's own definition (`amending = wasFinalized && !approved`), persisted to the DB in real time via `/save-summary`. So the gate reads current state without diffing anything:
  - First-time draft not yet approved → **block**
  - Amendment started, not re-approved (`approved` reset to `false`) → **block**
  - Recording made, never generated/approved → **block**
  - Approved (finalized / re-finalized) → **allow**
- "Scribe was used" excludes commands synced from the note body (`section_key == "from_the_note"` / the Additional Commands rail) — those aren't Scribe work product.
- Only the Lock transition (`state == "LKD"`) is gated; push-charges (`PSH`) is not.

## Testing

- 13 new unit tests in `tests/hyperscribe/scribe/handlers/test_note_lock_guard.py` covering state filtering, the usage predicate (including `from_the_note` exclusion and blank-section handling), approved short-circuit, and the configurable tab-name message.
- Full unit suite green (`tests/`: 2561 passed); ruff + mypy clean.
- Verified live on a Canvas instance: handler fires on `NOTE_STATE_CHANGE_EVENT_PRE_CREATE` and blocks/allows as expected.

### DIFF
--- a/hyperscribe/CANVAS_MANIFEST.json
+++ b/hyperscribe/CANVAS_MANIFEST.json
@@ -174,6 +174,15 @@
         }
       },
       {
+        "class": "hyperscribe.scribe.handlers.note_lock_guard:NoteLockGuard",
+        "description": "Block locking a Scribe-documented note until the Scribe tab is finalized",
+        "data_access": {
+          "event": "NOTE_STATE_CHANGE_EVENT_PRE_CREATE",
+          "read": [],
+          "write": []
+        }
+      },
+      {
         "class": "hyperscribe.scribe.print.button:PrintScribeNoteButton",
         "description": "Note header dropdown button that launches a print preview of the scribe summary",
         "data_access": {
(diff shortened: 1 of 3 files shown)


kevincareycanvas:
cerberus review

kevincareycanvas:
Completed UAT

PR #306, opened 2026-06-16
KOALA-5808: disable chart command buttons while Scribe tab is active

kevincareycanvas:
## What

Hides the patient-chart command buttons (Conditions, Medications, Allergies, Vitals, etc.) while the Scribe note tab is active, so providers chart through the Scribe summary instead of the legacy rail. Buttons are restored when the provider leaves the Scribe tab.

Buttons are set to **HIDDEN** (fully removed from the chart sections) while the Scribe tab is active.

## How

- **`ScribeApp.handle()`** emits a `ConfigureCommandButtons(... HIDDEN ...)` effect when the Scribe tab opens (covers default-open on note load).
- **`POST /scribe-session/configure-command-buttons`** (`ScribeSessionView`) toggles all 16 chart-section locations HIDDEN/VISIBLE for the note's patient; read-level auth.
- **`summary.js`** posts to that endpoint on `NOTE_TAB_CHANGE` — hide when the active tab is `hyperscribe__scribe`, restore otherwise.
- **`NoteCommandButtonsRestoreHandler`** (new, on `NOTE_CLOSED`) restores buttons when the note is collapsed/closed. Required because `ConfigureCommandButtons` is sticky and patient-scoped, and the Scribe iframe (the other restore path) is destroyed once you leave the note.
- Shared `configure_command_buttons_effect()` helper keeps the hide/restore effects symmetric across all three call sites.

## Install fix (resolved)

Initial installs to scribeqa-playground/sandbox returned HTTP 500. **Root cause (confirmed via infra/Sentry):** `plugin_io_plugin.version` is `varchar(64)` and the upload writes `plugin_version` verbatim; the generated string was 65 chars — `2026-06-15 v0.3.0 (kevin/koala-5808-head-patient-summary-buttons)` — the long branch suffix pushed it one over, raising `django.db.utils.DataError` on `PATCH /plugin-io/plugins/hyperscribe/`.

Fixed by shortening `plugin_version` to `2026-06-15 v0.3.0 (koala-5808)` (30 chars) and trimming `tags.version_branch` so a regenerated string stays under the limit.

(Platform follow-up, separate: the upload should validate version length and return a 400 instead of letting the overflow surface as a 500.)

## Runtime note

The feature uses two canvas SDK APIs introduced in **0.163.0** — the `ConfigureCommandButtons` effect and the `NOTE_CLOSED` event — so `sdk_version` is set to `0.163.0`. This was **not** the install blocker (that was the version string above), but the instance must run canvas >= 0.163.0 for the button hiding to actually function; worth confirming during UAT.

## Test plan

- `uv run pytest tests/hyperscribe` — 2098 passing.
- `uv run canvas validate-manifest hyperscribe` — valid.
- On a canvas >= 0.163.0 instance: install cleanly, open the Scribe tab → chart-section command buttons are hidden; switch tabs or close the note → buttons restored.


kevincareycanvas:
cerberus review

kevincareycanvas:
janus uat

kevincareycanvas:
janus uat

kevincareycanvas:
Manual UAT passed. Merging now.

PR #308, opened 2026-06-17
KOALA-5855: fix(scribe): accurate medication strength + blank placeholder sig

kevincareycanvas:
Jira: [KOALA-5855](https://canvasmedical.atlassian.net/browse/KOALA-5855)

Two related accuracy fixes in the **nabla/scribe** recommendation path (`hyperscribe/scribe/recommendations/`). The legacy `hyperscribe/commands/` path is untouched.

## 1. Wrong medication strength (20 mg → 10 mg)

**Problem:** Nabla accurately captures "Lisinopril 20 mg" but scribe recommends "Lisinopril 10 mg".

**Root cause:** The medication-statement and prescription recommenders resolved an FDB code by searching CanvasScience for a bare keyword and blindly taking `results[0]` — no strength matching, no selection step. The LLM correctly captured "20 mg" in `medication_name`, but the resolver discarded it and overwrote the display with whatever group FDB ranked first (commonly 10 mg).

**Fix:** New shared helper `_medication_match.resolve_medication_detail()` searches FDB with the full medication name first, then keywords, and deterministically selects the candidate whose description carries the **same normalized strength** (e.g. `20mg`), falling back to the first candidate only when nothing matches. Strength tokens are exact (`20mg` ≠ `120mg` / `2.5mg`); combination products handled. Both recommenders delegate to it (removing the duplicated buggy resolver). Extraction prompts reinforced to preserve the exact stated strength.

## 2. "<UNKNOWN>" in the SIG

**Problem:** Recommendations sometimes show literal `<UNKNOWN>` in the SIG line.

**Root cause:** The recommendation schema requires `sig` as a non-null `str`, so when a note states no directions the LLM fills the field with a placeholder. This is self-imposed by scribe — the SDK's `MedicationStatementCommand.sig` is `str | None`.

**Fix:** `sanitize_sig()` blanks any angle-bracket-wrapped value (`<UNKNOWN>`, `<none stated>`, …) and known no-value tokens (`unknown` / `n/a` / `none` / `tbd` / `-`). The scribe UI already hides an empty sig, and `build()` maps `"" → None` on the command, so a directionless medication now shows no sig line instead of placeholder text.

## Scope / risk
- Self-contained: resolvers have no external callers; registry wiring unchanged.
- Behavior note for QA: strength matching is blind to dosage **form/route**, so a tablet vs. liquid with the same numeric strength could differ from before — worth a quick validation. Combination-product strengths fall back to first-candidate (no regression). No-dose meds and not-found cases behave as before.
- Follow-up (not in this PR): making `sig: str | None` in the schema + "leave empty if not stated" prompt guidance would stop the placeholder at the source; the sanitizer covers it regardless.

## Tests / checks
- New `test_medication_match.py` (strength extraction/selection/resolution + `sanitize_sig`), regression cases in both recommender test suites (20 mg resolves to 20 mg group even when FDB returns 10 mg first; `<UNKNOWN>` sig surfaces blank).
- mypy clean; ruff check/format clean; pytest scribe recommendations + commands green.

[KOALA-5855]: https://canvasmedical.atlassian.net/browse/KOALA-5855?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/hyperscribe/scribe/recommendations/_medication_match.py
+++ b/hyperscribe/scribe/recommendations/_medication_match.py
@@ -34,12 +34,14 @@ import re
 from hyperscribe.libraries.canvas_science import CanvasScience
 from hyperscribe.structures.medication_detail import MedicationDetail
 
-# A strength is a number (optionally decimal, optionally a hyphen/slash-joined
-# compound like "20-12.5" for combination products) followed by a dosage unit.
-# We capture the whole "<number><unit>" token so matching is exact and "20 mg"
+# A strength is a number followed by a dosage unit. The number may carry comma
+# thousands separators ("1,000") and may be a hyphen/slash-joined compound that
+# shares one unit ("20-12.5 mg", "160/4.5 mcg") for combination products. Each
+# component is normalized to an exact "<number><unit>" token so that "20 mg"
 # never matches "120 mg" or "2.5 mg".
+_NUMBER = r"\d[\d,]*(?:\.\d+)?"
 _STRENGTH_RE = re.compile(
-    r"(\d+(?:\.\d+)?(?:\s*[-/]\s*\d+(?:\.\d+)?)*)\s*"
+    rf"({_NUMBER}(?:\s*[-/]\s*{_NUMBER})*)\s*"
     r"(mcg|mg|g|ml|%|iu|meq|units?)\b",
     re.IGNORECASE,
 )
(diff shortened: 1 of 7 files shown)


kevincareycanvas:
cerberus review

PR #309, opened 2026-06-18
feat(scribe): generic specialty referrals with validated indication

kevincareycanvas:
## Summary
Makes scribe **referral** recommendations generic to the specialty and commit-ready, in the nabla/scribe path (`hyperscribe/scribe/`). Legacy `hyperscribe/commands/` is untouched.

## Changes
1. **Generic referrals only.** `ReferRecommender` no longer runs the science-service provider lookup; every referral is recommended to the specialty (no specific provider). The manual provider-search endpoint (`search_refer_providers`) is unchanged and still available in the UI.
2. **Validated indication (diagnosis code).** The LLM also extracts the condition the referral addresses (`indication`), and a new pure helper `link_referral_diagnoses` attaches an ICD-10 code **already present in the note** — diagnose commands → science-validated diagnosis suggestions → unmatched conditions. Never fabricates: no match leaves `diagnosis_codes` empty.
3. **Insert/sign-ready.** A refer command auto-signs on insert and Canvas core requires a recipient + clinical question + notes + indication. So generic referrals now carry: a placeholder recipient (non-blank `first_name`/`last_name` = "TBD", `practice_name` = specialty — Canvas core rejects a blank `first_name`), a default `clinical_question` ("Assistance with Ongoing Management"), and `notes_to_specialist` (reason → indication → "Referral to {specialty}").
4. **`ReferParser.validate` tightened** to require all four sign fields, so an incomplete referral fails loudly at `/insert-commands` instead of silently rolling back at sign (same guard `_rx_validation` provides for prescribe).
5. **Priority dropped** from the referral pipeline (schema + prompt + data) — left blank for the provider to set.
6. PHI-safe logging in the refer recommender (no raw LLM response on error/parse paths).

## Tests / checks
- Rewrote `test_refer.py` for generic-only + placeholder + defaults; added `test_referral_diagnosis.py` (match priority, conservative matching, no-fabrication); added `ReferParser.validate` tests; updated `test_builder.py` to the 4-field contract.
- Full scribe suite green (1,229), ruff + mypy clean.

## Verified live
Confirmed in a real instance that generic referrals now originate + sign into the note body (the blank-`first_name` ORIGINATE failure is resolved).

## Notes / out of scope
- The standalone `/recommend-commands` endpoint returns generic referrals without linked diagnosis codes (no diagnosis context there); linking runs in the generate-summary flow.
- `clinical_question` default is a neutral guess when the AI doesn't infer one — provider can adjust.
- Diagnosis matching is conservative (exact/substring) to avoid mis-linking.

### DIFF
--- a/hyperscribe/scribe/api/session_view.py
+++ b/hyperscribe/scribe/api/session_view.py
@@ -81,6 +81,7 @@ from hyperscribe.scribe.commands.problem_list_match import (
     prefer_patient_specific_codes,
 )
 from hyperscribe.scribe.recommendations import recommend_commands
+from hyperscribe.scribe.recommendations._referral_diagnosis import link_referral_diagnoses
 from hyperscribe.scribe.recommendations.diagnosis_suggestion import suggest_diagnoses
 from hyperscribe.scribe.recommendations.interactions import (
     check_recommendation_interactions,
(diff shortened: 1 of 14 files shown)


kevincareycanvas:
cerberus review

PR #311, opened 2026-06-25
feat(scribe): Manual Mode Only setting + visit-type selector polish

kevincareycanvas:
## Summary

Adds a new plugin setting to restrict Scribe to manual documentation, plus a few related manual-mode and UI refinements.

- **`ScribeManualModeOnly` setting** — when the secret is truthy, the **Start AI Scribe** button is hidden so only **Manual** can be selected. Default off → behavior unchanged. Threaded through the existing `alertFacilityEnabled` feature-flag path (manifest secret → `scribe_view` context → `index.html` config → `App` → `Scribe`), so no new plumbing concepts.
- **Tidy empty manual-mode sections** — stop pre-seeding an empty Plan card under Assessment & Plan (added on demand via **+ Plan**, mirroring the existing Vitals exclusion); hide the **Social History** section when empty (it has no manual input affordance).
- **Fix: restore Add/Resolve Condition in an empty A&P** — removing the pre-seeded Plan card left `assessment_and_plan` with no commands, which tripped a `!cmds` render guard and dropped the section's Add/Resolve Condition buttons. Plan sections with condition-add affordances are now exempt from that guard, so the buttons render without re-introducing the empty card.
- **Visit-type selector polish** — replace the cramped OS default `<select>` arrow with a custom inset chevron (`appearance:none` + inline SVG), add a hover border and navy focus ring, and align the selector's corner radius (6px) and height with the Start AI Scribe / Manual buttons.

## Plugin version
`2026-06-24 v0.3.13`

## Testing
- Python suite: **2,623 passed**.
- Caveat: all changes here are frontend (the flag plumbing, `soap-group.js`/`summary.js` logic, `styles.css`) and the repo has no JS test harness, so the green Python suite confirms nothing else regressed but does **not** cover these changes. Recommend a manual QA pass of the manual-mode A&P flow (add condition, resolve condition, add plan, finalize) and the selector appearance.

### DIFF
--- a/hyperscribe/CANVAS_MANIFEST.json
+++ b/hyperscribe/CANVAS_MANIFEST.json
@@ -1,6 +1,6 @@
 {
   "sdk_version": "0.155.0",
-  "plugin_version": "2026-06-24 v0.3.12",
+  "plugin_version": "2026-06-24 v0.3.13",
   "name": "hyperscribe",
   "description": "Create commands based on the content of an audio discussion between a patient and a provider.",
   "url_permissions": [
--- a/hyperscribe/CANVAS_MANIFEST.json
+++ b/hyperscribe/CANVAS_MANIFEST.json
@@ -1,6 +1,6 @@
 {
   "sdk_version": "0.155.0",
-  "plugin_version": "2026-06-24 v0.3.11",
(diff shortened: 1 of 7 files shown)


kevincareycanvas:
cerberus review

PR #316, opened 2026-06-29
feat(scribe): polish A&P condition card UI

kevincareycanvas:
Polishes the diagnose/assess condition cards in the Scribe Summary tab (Assessment & Plan) for a cleaner, more cohesive, lower-training experience. Iterated on with Kevin via interactive mockups.

Targets `feat/accept-reject-workflow-upgrade` and is merged on top of that branch's accept/reject redesign (see "Merge notes" below).

## What changed

**Collapsed (read) view**
- Removed the Background help text and character counter from the read view.
- Hide the Background section entirely when there is no background (no label, no "No background text", no counter).

**Edit view**
- Replaced the verbose gray hint with a concise "Carries forward to future notes" line under the Background box.
- When there's no preexisting background, show a "+ Add background" disclosure (reusing the refer card's `.refer-add-pill` pattern) instead of an empty field.
- Tightened vertical rhythm: tight within a field group, generous between groups.
- Character count shows only in edit mode; over limit the numerator is bold red and the denominator stays gray (scoped so legacy counters elsewhere are untouched).

**No-code state, manual search, and "Change diagnosis" — unified into one integrated picker**
- Replaced the always-open search + stacked suggestion buttons with a single picker: search field on top, recommendations / live search results listed below. The diagnosis name leads; the ICD code is a quiet fixed-width identifier.
- The same picker powers the no-code state and the header "Change" flow.

**Over-limit handling**
- Surfaces "Background too long" / "Assessment too long" warning pills in the actions column (mirroring "Missing Diagnosis Code"), including read-only rows so the signal is never hidden.

Applied to both diagnose (`rec-diagnose`) and matched-condition (`rec-assess`) cards. Preserves all KOALA_5635 safeguards (clear-on-ICD-change, charge-pointer safety, background round-trip). Two `test_session_view.py` pins updated to match the new help-text/counter behavior.

## Merge notes (vs the accept/reject redesign on the base branch)
This branch's UI touches the same actions column the base branch redesigned. Resolved by keeping the base branch's work and layering this on top:
- `diagnose-row.js`: kept the base branch's `aiPending` prop + `AI_SPARKLE` "needs review" sparkle in the title.
- `soap-group.js`: kept the base branch's `renderRecActions(...)` accept/reject UI and `rec-remove-x` button; the over-limit warning pills are rendered in front of it.

## Verification
- `node --check` clean on both JS files.
- Full pre-commit gate passes (ruff, ruff format, mypy, pytest — all tests).
- Two prior high-effort code-review passes; all confirmed regressions fixed.
- Component behavior verified in a local harness importing the real module across every state (no-code picker, search, select, Change, collapsed with/without background, edit, over-limit).

### DIFF
--- a/hyperscribe/scribe/static/diagnose-row.js
+++ b/hyperscribe/scribe/static/diagnose-row.js
@@ -284,12 +284,10 @@ export function DiagnoseRow({ command, commandIndex, onEdit, onDelete, readOnly,
 
   const conditionHeader = data.condition_header || command.display;
   const formattedCode = hasCode ? formatIcdCode(data.icd10_code) : null;
-  // Name leads, the ICD code trails it as a quiet gray identifier (matches the
-  // name's size/weight, color-differentiated only). Changing the code is done
-  // via the header "Change" affordance, not by clicking the code.
-  const title = hasCode
-    ? html`${data.icd10_display || conditionHeader}<span class="diagnose-icd-code">${formattedCode}</span>`
-    : conditionHeader;
+  // Name leads; the ICD code trails it as a small green pill (a sibling of the
+  // title so the pill aligns cleanly). Changing the code is done via the header
+  // "Change" affordance, not by clicking the code.
+  const titleText = hasCode ? (data.icd10_display || conditionHeader) : conditionHeader;
 
   const recCodes = suggestions || [];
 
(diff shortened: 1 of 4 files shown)


kevincareycanvas:
cerberus review

kevincareycanvas:
cerberus review

kevincareycanvas:
Re: the Cerberus finding on `styles.css:544` (`:has(.editing)` hiding `.recommendation-actions` during text editing) — **intended, leaving as-is.**

When you edit a condition card's text, the row's edit area carries the `editing` class, so the right-hand actions column (Accept/Reject + the over-limit "too long" pills) hides until Save/Cancel. This is deliberate:

- It matches how the matched-condition (assess) cards already behave — adding the `editing` class to the diagnose card makes both consistent.
- The edit view has its own Save/Cancel, so showing Accept/Reject simultaneously would be redundant.
- No over-limit signal is lost: while editing, the live character counter (bold-red numerator) shows and Save is disabled; the "too long" pill reappears in the collapsed view after Save/Cancel.

No code change for this one.

PR #317, opened 2026-06-30
fix(scribe): always surface Physical Exam in the summary tab

kevincareycanvas:
## Problem

When the Nabla payload contained no `physical_exam` content, the Scribe Summary tab dropped the **Physical Exam section entirely** — the render branch (`soap-group.js`) only drew the card when a command already existed. There was no way to document a physical exam that Nabla didn't capture.

## Fix

Render the Physical Exam on **every visit type**, even when Nabla omits it:
- **Documented card** when a PE command exists.
- **Empty "click to add" editor** otherwise; the provider's first Save routes through `handleAddPhysicalExam`, which promotes the draft into a real, index-addressable command.
- An **untouched empty exam is dropped by the existing `insertable` filter**, so nothing blank is ever written to the chart.

`ENSURE_KEYS` now always ensures the `physical_exam` section so the render branch is reached even when Nabla returns nothing.

The psychiatry-only **Mental Status Exam** gets the same always-present treatment (empty editor when omitted), and the empty-state copy is simplified to `No <section> documented.`

## Scope / behavior

| | Non-psych | Psychiatry |
|---|---|---|
| Physical Exam | Always present (card or empty editor) | Same |
| Mental Status Exam | Not shown | Always present (card or empty editor) |

Physical Exam behaves **identically across all visit types** — no template-specific add/remove logic.

Files (frontend only):
- `summary.js` — `handleAddPhysicalExam`/`handleAddMentalStatusExam`, `ENSURE_KEYS`, section order
- `soap-group.js` — PE/MSE render branches
- `sections-exam-row.js` — empty-state copy

## Notes for reviewers

- These are Preact/htm view components; the repo has **no JS test harness**, so this rests on manual verification. The full Python `pytest` suite still passes (pre-commit), and no Python touched.
- The copy change (" Click to add." removed) is in the shared `ExamSectionsRow`, so it also affects the **ROS** and **Mental Status Exam** empty states — intentional and consistent.

## Verification

Manually verified against a psych transcript with no physical-exam content: PE renders as an empty editor that can be documented, and an untouched empty exam is not written to the chart.

### DIFF
--- a/hyperscribe/scribe/static/sections-exam-row.js
+++ b/hyperscribe/scribe/static/sections-exam-row.js
@@ -146,7 +146,7 @@ export function ExamSectionsRow({
     return html`
       <div class=${`exam-rows${readOnly ? '' : ' exam-clickable'}`} onClick=${enterEdit}>
         ${sections.length === 0
-          ? html`<div class="exam-empty">No ${LABEL[sectionKind]} documented.${readOnly ? '' : ' Click to add.'}</div>`
+          ? html`<div class="exam-empty">No ${LABEL[sectionKind]} documented.</div>`
           : sections.map((s, i) => html`
             <div class="exam-row" key=${s.key || i}>
               <div class="exam-sys">${s.title}</div>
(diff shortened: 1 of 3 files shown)


kevincareycanvas:
cerberus review

PR #325, opened 2026-07-16
fix(scribe): resilient, lossless Canvas Scribe transcript on connectivity loss

nmpsilva:
## Ticket
[KOALA-5934](https://canvasmedical.atlassian.net/browse/KOALA-5934)

## What
On real visits, connectivity loss during a Canvas Scribe session corrupted the transcript — duplicated/garbled content, a multi-minute dead gap, all-Unspecified speakers — and the note was then generated from that corrupted transcript. This makes the realtime (websocket) path resilient to connectivity loss and, at Finish, **lossless**: no captured audio is silently dropped.

## Implementation
Two layers.

**Transport hardening.** Transcript items that re-arrive across a reconnect are de-duplicated, and internal reconnects are anchored to audio-time instead of wall-clock so replayed audio keeps its true timeline position — together this eliminates both the duplication and the dead gap. The reconnect path was rewritten so a socket that opens then immediately drops backs off exponentially and trips a circuit breaker instead of storming, and disconnects/reconnects are instrumented in the audit log. In-flight audio is capped at ~3s to stay clear of Nabla's backpressure limit.

**Lossless finish.** Instead of force-closing after a fixed wait (which truncated the tail — the socket's own 30s end-timeout was closing it mid-drain, the real cause of the earlier data loss), the client keeps draining buffered audio to the service after Finish and only finalizes — and generates the note — when the buffer is fully drained **or** the provider explicitly accepts a gap via "Finalize now", recorded in a `FINISH_DRAIN` audit event. Because transcription runs at ~realtime, a long outage means the note finalizes after a short delay rather than at Finish; that delay is the accepted cost of losing nothing.

Mic echo-cancellation is also made server-configurable (default unchanged) for two-speaker room capture.

[KOALA-5934]: https://canvasmedical.atlassian.net/browse/KOALA-5934?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


nmpsilva:
Janus uat

PR #330, opened 2026-07-22
feat(scribe): extend Alert Facility to prescribe, adjust prescription, refill

kevincareycanvas:
## Summary

Extends **Alert Facility** to the **prescribe**, **adjust prescription**, and **refill** commands, and — through review — hardens the whole feature across all five supported commands (those three plus **medication statement** and **stop medication**): a clean display treatment, per-command-type defaults, a per-command allow-list secret, and strict backwards-compatibility so historical notes are never altered.

This is the **producer** side; the companion validation plugin ([Medical-Software-Foundation/canvas#417](https://github.com/Medical-Software-Foundation/canvas/pull/417)) enforces the field on the Canvas side for these schema keys.

## What's included

**1. Extend to prescribe / adjust prescription / refill**
- `pending_metadata` added to the three Rx parsers; `build_effects` emits the `alert_facility` command metadata via the existing generic path.

**2. Display treatment**
- Replaced the amber full-width pill with a neutral `Alert Facility: Yes/No` line matching the order details line. Shown consistently in recommendation, documented, and committed states (incl. the flat ADDITIONAL COMMANDS cards, read back from `CommandMetadata`). Fixed stop_medication rendering the flag on the right.

**3. Per-command-type defaults**
- Prescribe & adjust prescription default **Yes**; refill, medication statement & stop medication default **No**. Applies to the toggle's starting position, the value recorded at commit, and freshly generated cards (so a new card shows its default without being opened). An explicit stored value always wins.

**4. Per-command gating via a plugin secret**
- Replaced the global boolean `AlertFacilityEnabled` with **`AlertFacilityCommands`** — a comma-separated allow-list of command types (matching the `ScribeNoteTypes` convention). A command shows the toggle only when its `command_type` is in the list; empty/unset = off everywhere. New shared module `hyperscribe/scribe/commands/_alert_facility.py` is the single source of truth (supported types, schema-key map, per-type defaults, parser).

**5. Backwards compatibility (historical notes untouched)**
- Display shows a value **only when one is genuinely recorded** — never a fabricated default. For committed/locked notes the source of truth is the committed `CommandMetadata`: `_load_summary` reconciles the cache against it, so a pre-feature command with no committed value shows **no** line, while a command with a committed value shows it. `handleSave` no longer fabricates a value on unrelated edits.

**6. Review hardening — make the reconcile invariant complete & uniform**
- **Amend re-emits metadata.** `build_amend_edit_effects` now takes `feature_flags` and re-emits `pending_metadata` for the void+recreated command (ordered originate -> metadata -> commit, mirroring `build_effects`). Previously an amended `medication_statement` / `stop_medication` committed with no `alert_facility` row, so both read paths treated it as pre-feature and **silently dropped a genuinely-recorded value**. No-op for types without pending metadata or when gated off.
- **Recommendations reconciled symmetrically.** `_inject` stamps a per-type default onto recommendations as well as commands, but `_load_summary` only reconciled commands. It now reconciles **both** surfaces, so a committed recommendation shows its true committed value (or nothing when pre-feature); unaccepted recommendations have no committed `command_uuid` and keep their in-flight default.

## Behavior matrix
- **New card** -> shows its per-type default (even unopened); records that value at commit.
- **Explicitly toggled** -> shows/records the chosen value.
- **Committed + no metadata (pre-feature / locked)** -> no Alert Facility line.
- **Command type not in `AlertFacilityCommands`** -> no toggle, no line, no record.

## Migration
`AlertFacilityEnabled` is removed. Set `AlertFacilityCommands` in every environment to the desired command types (e.g. `prescribe,adjust_prescription,refill,medication_statement,stop_medication` for all on). The old secret has no effect.

## Tests
- Parser `pending_metadata` (per-type default / explicit / allow-set gating) for all five commands.
- `build_effects` inline metadata emission.
- `_alert_facility` parser/maps; generation-time default injection; `/note-commands` allow-list + metadata gating.
- `_load_summary` committed-record reconciliation (set / strip / preserve / add; empty + disallowed no-ops), now also covering the **recommendations** surface.
- Amend re-emits `alert_facility` metadata for the recreated command (exact `UPSERT_COMMAND_METADATA` payload) + gating (empty allow-set / no `feature_flags` -> no metadata effect).
- Frontend source-pin tests for the display gates, touched-gated save, and per-command `.has()` gating.
- Full suite green — 2918 passed (ruff + mypy + pytest via pre-commit).

### DIFF
--- a/hyperscribe/CANVAS_MANIFEST.json
+++ b/hyperscribe/CANVAS_MANIFEST.json
@@ -260,7 +260,7 @@
     "views": []
   },
   "secrets": [
-    "AlertFacilityEnabled",
+    "AlertFacilityCommands",
     "AnthropicAPIKey",
     "APISigningKey",
     "AudioHost",
--- a/hyperscribe/libraries/constants.py
+++ b/hyperscribe/libraries/constants.py
@@ -127,7 +127,7 @@ class Constants:
     SECRET_NOTION_FEEDBACK_DATABASE_ID = "NotionFeedbackDatabaseId"
     SECRET_SCRIBE_BACKEND = "ScribeBackend"  # JSON: {"vendor": "nabla", "region": "us", ...}
     SECRET_SCRIBE_NOTE_TYPES = "ScribeNoteTypes"  # Comma-separated note type names
-    SECRET_ALERT_FACILITY_ENABLED = "AlertFacilityEnabled"
+    SECRET_ALERT_FACILITY_COMMANDS = "AlertFacilityCommands"  # Comma-separated command types
     SECRET_SCRIBE_DEBUG_STAFFERS = "ScribeDebugStaffers"  # Comma-separated staff keys for debug apps
     SECRET_VISIT_TEMPLATES = "VisitTemplates"  # JSON: {"templates": [...]}
     SECRET_NOTION_API_KEY = "NotionAPIKey"
--- a/hyperscribe/scribe/api/scribe_view.py
+++ b/hyperscribe/scribe/api/scribe_view.py
@@ -12,6 +12,7 @@ from canvas_sdk.v1.data.staff import Staff
 from hyperscribe.libraries.helper import Helper
 from hyperscribe.models.scribe import ScribeTranscript
 from hyperscribe.scribe.api.session_view import _load_initial_data
+from hyperscribe.scribe.commands._alert_facility import parse_alert_facility_commands
 
 
 def _safe_json(data: Any) -> str:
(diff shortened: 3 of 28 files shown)


kevincareycanvas:
cerberus review

kevincareycanvas:
janus uat

kevincareycanvas:
cerberus review

kevincareycanvas:
janus uat

kevincareycanvas:
janus uat

I have configured the appropriate plugin secret so the facility toggle changes are testable

PR #331, opened 2026-07-27
feat(scribe): retire accept/reject on condition cards — pick a code to accept, ✕ to dismiss

kevincareycanvas:
## What & why

The **Reject / Accept** pair on A&P condition cards was redundant, and provably so:

- `handleSelect` already set `accepted: true` the moment a code was picked from the picker.
- `handleEdit` force-set `accepted = false` whenever there was no `icd10_code`.

Together those meant the **Accept button on an uncoded card was permanently disabled and could never be clicked**. Its only reachable use was restoring a rejected card — so the pair cost ~180px of the card header and added a decision the picker already satisfied.

**New model:** selecting a diagnosis from the list of options *is* the accept; a single ✕ in the upper right *is* the reject.

Design was iterated over three rounds against a rendered high-fidelity mockup before any product code was touched. The mockup was a throwaway artifact and is not included here.

**Scope: condition cards only** (`diagnose` + `assess`). Medications, allergies, prescriptions, referrals, tasks and orders keep `Reject`/`Accept` and `renderRecActions` **untouched**.

## Behaviour

- `accepted` is now fully **derived** (`coded && !dismissed`). Still written, because the pre-insert filter, the diagnose→assess flip and the print filter all read it.
- ✕ marks `data.rejected`; the existing top-bar toggle governs visibility (relabelled **"Hide Dismissed Recommendations"**).
- A dismissed card is dimmed, carries **no controls**, and is itself the restore target — rendered as a real focusable button (`role="button"`, `tabindex="0"`, Enter/Space). That is a correctness requirement, not polish: it is the only route back.
- **Assess ✕ now dismisses recoverably instead of deleting the row.** The two condition-card types show an identical ✕ in an identical place, so they must not differ in permanence — an assess card carries AI-written assessment text a mis-click shouldn't destroy.
- Picker ⋮ menu gains **"Keep current diagnosis"**, offered only when re-coding an already-coded card. It replaces the picker's own ✕, so exactly one ✕ appears on a card and it unambiguously means "dismiss this condition". Escape still cancels.
- **Sign gate split**: conditions are gated on *"has a code been picked?"* (`uncodedConditionCount`) rather than folded into the accept/reject undecided count. Deliberately not just dropped — the uncoded filter in `handleInsert` only removes such a card from `insertable`, which without a gate would be **silent data loss**.

## Visual

- Card header now shares **one 24px line box**: title line-height, every chip and the ✕ all adopt it, and chips pin to the title's **first** line. Previously the ✕ sat **7.5px below** the title (21px line box vs a 4px-padded 28px button), and a wrapped title dragged the chips to the block centre while the ✕ stayed high.
- Action column pulled **6px right** so the 15px *glyph* — not the 28px hit target — is 17.5px from both card borders: on the corner diagonal, and flush with the picker column. Hit target still stops 11px from the border, so nothing is clipped.
- Amber `Missing: Diagnosis Code` pill → **`Needs Diagnosis Code` chip** mirroring the green ICD-10 chip on *every* computed property. A card now shows a green code chip or an amber needs-code chip in the same slot at the same size, so coded/uncoded is scannable down the column.
- Title→picker gap **24px → 12px**; the 18px below is kept so the asymmetry groups the picker with the title it belongs to.
- Picker's search-row separator moved onto the body content, fixing a stray gray line across the rounded corner on cards with no candidate codes.

## Bugs fixed

1. **Dismissed conditions were unrecoverable.** The visibility toggle only rendered when `recommendations.length > 0`, and that array excludes condition cards (they live in `commands`). With `hideRejected` defaulting to `true`, a note whose only recommendations were conditions showed **no toggle at all**. Survivable only while the card carried its own Accept button — removing that turned it into data-shaped loss.
2. **Dismissed conditions stayed linkable to charges.** The three charge-matrix predicates gate on `accepted`, and for `assess` the test is `accepted !== false` — assess data has no `accepted` field, so `undefined` passed. All three now also check `!data.rejected`. (Found during implementation, not in the plan.)
3. **Assess headers started 4px below diagnose headers.** `.diagnose-row`'s `4px 0` padding only ever reached assess cards, since `.rec-diagnose` dissolves that element with `display: contents`.

## Note for reviewers

`.rec-warning-pill` was **not** restyled — it has 7 other live uses (`Background too long`, `Assessment too long`, `Missing: X` on the other card families) which keep the louder bordered pill, since those are genuine problems rather than a routine "not picked yet". The new chip is a separate class.

## Testing

- `1483` tests pass unchanged (`tests/hyperscribe/scribe/`). Behaviour is frontend-only, so a failure there would have meant the data contract moved.
- Geometry verified by **measuring rendered layout against the shipped stylesheet** — ✕ and chip centres vs the title's first line, glyph insets, chip parity — rather than by reading CSS. That caught two of the offsets above that reasoning had missed.

⚠️ **Not yet exercised against a live Canvas instance.** There is no JS test harness for this UI, so the interaction behaviour is unverified. Worth walking before merge — especially **(4)** and **(6)**, which are the assess behaviour change and the charge-matrix fix that wasn't in the original plan:

1. Pick a code → card codes *and* settles in one action
2. `✎ Change` → ⋮ offers *Move to Wrap Up* + *Keep current diagnosis*; no picker ✕
3. ✕ dismiss → toggle off → **Tab to the card and press Enter** to restore
4. Same ✕/restore on an **assess** (existing-condition) card
5. Sign gate: uncoded condition blocks with the new copy; a dismissed one does not
6. **Charges panel: a dismissed condition disappears from the linkable list**

Plugin version bumped to `2026-07-27 v0.4.41`.

### DIFF
--- a/hyperscribe/CANVAS_MANIFEST.json
+++ b/hyperscribe/CANVAS_MANIFEST.json
@@ -1,6 +1,6 @@
 {
   "sdk_version": "0.155.0",
-  "plugin_version": "2026-07-07 v0.4.40",
+  "plugin_version": "2026-07-27 v0.4.41",
   "name": "hyperscribe",
   "description": "Create commands based on the content of an audio discussion between a patient and a provider.",
   "url_permissions": [
--- a/hyperscribe/scribe/static/diagnose-row.js
+++ b/hyperscribe/scribe/static/diagnose-row.js
@@ -6,7 +6,10 @@ const html = htm.bind(h);
 
 const ICON_PENCIL = html`<svg viewBox="0 0 24 24" fill="none" stroke="currentColor" stroke-width="2" stroke-linecap="round" stroke-linejoin="round"><path d="M12 20h9"/><path d="M16.5 3.5a2.1 2.1 0 0 1 3 3L7 19l-4 1 1-4Z"/></svg>`;
 const ICON_SEARCH = html`<svg viewBox="0 0 24 24" fill="none" stroke="currentColor" stroke-width="2" stroke-linecap="round" stroke-linejoin="round"><circle cx="11" cy="11" r="7"/><line x1="20" y1="20" x2="16.65" y2="16.65"/></svg>`;
-const ICON_X = html`<svg viewBox="0 0 24 24" fill="none" stroke="currentColor" stroke-width="2" stroke-linecap="round"><line x1="6" y1="6" x2="18" y2="18"/><line x1="6" y1="18" x2="18" y2="6"/></svg>`;
+// Revert arrow for the picker menu's "Keep current diagnosis" item. (The picker's own ✕
+// close button is gone — that menu item replaces it, so there is exactly one ✕ on a card
+// and it unambiguously means "dismiss this condition".)
+const ICON_REVERT = html`<svg viewBox="0 0 24 24" fill="none" stroke="currentColor" stroke-width="2" stroke-linecap="round" stroke-linejoin="round"><path d="M9 14 4 9l5-5"/><path d="M4 9h11a5 5 0 0 1 0 10h-3"/></svg>`;
 // Vertical ⋮ for the search-bar overflow menu, and a "move to a list/section" glyph for its item.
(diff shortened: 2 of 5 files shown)


kevincareycanvas:
cerberus review

kevincareycanvas:
janus uat

kevincareycanvas:
Found a bug at time of manual UAT where changing a diagnosis code. Will file a ticket for a follow-up PR.

kevincareycanvas:
cerberus review

PR #332, opened 2026-07-28
KOALA-5808: restore chart command buttons after Confirm & Add to note

kevincareycanvas:
Follow-up to #306. That PR merged before this fix was pushed, so it landed on the branch but never made it into `feat/canvas-scribe`.

## The bug

After the provider clicks **Confirm & Add to note**, the chart-section command buttons stay hidden. Recovery requires a Scribe → Note tab round-trip or a page reload. This is the primary Scribe workflow, so it hit on every approval.

## Why the existing restore didn't cover it

`CLOSE_MODAL` *does* switch the note back to its body tab (`NoteTabsContext.onCloseApplication` → `setActiveIndex(applications.length)`), which fires a `NOTE_TAB_CHANGE` with `tab='note'`. We never receive it:

1. home-app clears the app's `frameData` in the same commit (`NoteApplicationIframe.onCloseMessage`).
2. The iframe's ref cleanup nulls the `MessageChannel` port (`useMessageChannelBroker`) during the **mutation** phase.
3. The `NOTE_TAB_CHANGE` broadcast runs in a **passive effect**, i.e. after that.

So `sendMessage` is a silent no-op by the time the tab change is broadcast, and the restore in `summary.js`'s `NOTE_TAB_CHANGE` handler never runs.

## The fix

POST the restore explicitly, immediately before the `CLOSE_MODAL` postMessage.

`await`ed rather than fire-and-forget: once `CLOSE_MODAL` lands the component unmounts and an in-flight fetch would be cancelled. Failures are logged and never block the close — the sign and tab-switch paths both re-assert visibility.

## Testing

- `test_summary_js_restores_command_buttons_before_close_modal` — follows the existing structural-static pin convention for `summary.js` in that module (marker comment + an assertion that the restore POST precedes the `CLOSE_MODAL` postMessage, since ordering is the whole point).
- I verified the pin **fails** against the pre-fix source rather than passing vacuously.
- 2899 Python tests pass; 18 JS tests pass.

## Caveat

The pin is structural, not behavioural — it does not execute the React code and cannot prove the fetch resolves before unmount. Worth confirming on a live instance: approve a Scribe summary and check the chart rail comes back without a tab round-trip.

## Related

- Remaining known limitations are tracked in KOALA-6545 (visibility leaks across users; last-writer-wins across plugins) and KOALA-6546 (refill approve/deny and unmapped review buttons can't be hidden at all).
- Manual note collapse still strands the buttons; that needs a home-app change and is documented in `command_buttons.py`.

### DIFF
--- a/hyperscribe/scribe/static/summary.js
+++ b/hyperscribe/scribe/static/summary.js
@@ -3045,6 +3045,33 @@ export function Scribe({ noteId, patientId, staffId, staffName, providerName, pr
         //     console.error('Failed to sign note:', signErr);
         //   }
         // }
+        // RESTORE_BUTTONS_BEFORE_CLOSE_MODAL
+        // Restore the chart-section command buttons BEFORE closing the modal.
+        // CLOSE_MODAL does switch the note back to its body tab, which fires a
+        // NOTE_TAB_CHANGE with tab='note' — but we never receive it. home-app
+        // clears the app's frameData and switches the tab in the same commit
+        // (NoteApplicationIframe.onCloseMessage -> NoteTabsContext
+        // .onCloseApplication), and the iframe's ref cleanup nulls the
+        // MessageChannel port during the mutation phase, before the passive
+        // effect that broadcasts the tab change. sendMessage is a silent no-op
+        // by then, so the restore in the NOTE_TAB_CHANGE handler above never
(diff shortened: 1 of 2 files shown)


kevincareycanvas:
cerberus review

kevincareycanvas:
successful UAT!

PR #334, opened 2026-07-29
fix(scribe): dismissing a duplicate condition card no longer blocks approval (KOALA-6555)

kevincareycanvas:
[KOALA-6555](https://canvasmedical.atlassian.net/browse/KOALA-6555) — P0 regression. Pylon #31325, #31372.

## The bug

Dismissing a duplicate Assessment card made the note **permanently unsignable**. Approve hard-blocked with the generic _"This command has invalid values. Open it to fix before approving."_ — naming a card the provider **could not see**, since `hideRejected` defaults to `true`. `NoteLockGuard` gates the `LKD` transition on `ScribeSummary.approved`, so signing was unreachable. 11 notes across 5 providers on brigade; one provider retried 8 times in ~12 minutes, another note logged 14 attempts.

Dismissing is currently the *only* way to remove an unwanted condition card, so the single action available to a provider facing a duplicate was the action that broke the note.

## Root cause — an asymmetry with `diagnose`, not a missing guard

`fd595251` gave condition cards a ✕ that marks `data.rejected` instead of deleting the row, and added a matching gate to the `insertable` filter so a dismissed card never reaches the chart. **That part was correct.** But it created a *third* kind of intentional drop, and neither consumer that interprets `dropped` was taught about it — the `COMMANDS_FILTERED` audit reason and `droppedForValidation`, which halts Approve. Both hand-rolled the same three-way classification, and they drifted.

Line `2706`'s trailing `&& !(c.data && c.data.rejected)` is a double negative, so a dismissed **diagnose** *passes* the gate, stays in `insertable`, never enters `dropped`, and is stripped later at `2786`. The new **assess** gate was a hard `return false` with no equivalent late strip, so it landed in `dropped` and read as a silent validation failure.

Verified against the stuck record's exact shape:

| | pre-fix | post-fix |
|---|---|---|
| `dropReason` | `validation` | `dismissed` |
| blocks Approve | **yes** | no |

## The fix

Extract the classification into **`command-drop.js`** (zero imports, so it is node-importable and can carry real unit tests — `summary.js` cannot be imported under Node) and have both consumers read `dropReason` / `isIntentionalDrop` from it. A new drop reason can no longer be added to one site without the other.

The exemption is scoped to the two condition families rather than a blanket `!c.data?.rejected`. Suppressing a validation error is only sound for a type that *also* has a matching "rejected ⇒ not insertable" gate; a blanket test would hand any future type that starts writing `data.rejected` silent error suppression without the exclusion — the lost-prescription mode `droppedForValidation` exists to catch. There is a test pinning that scoping.

**Two related fixes from the same root cause:**

- `codedNoteDiagnoses` fed dismissed-but-coded cards to the approve-time referral linker, so a referral could be committed citing a diagnosis that never gets inserted — and via the uniqueness test in `resolveReferralIndication`, a dismissed near-duplicate could make a genuine match look ambiguous and suppress a link that *should* have been made.
- `insertableCount` over-counted dismissed assess rows in the "Inserting N commands into note…" label.

## Deliberate deviations from the ticket's suggested fix

- **No `assess` branch in `_commandValidationReason`.** It would be dead code: assess can only fail two gates, and `droppedForValidation` already excludes the other one (`!display`).
- **Scoped, not blanket** exemption — see above.

## Ticket claims that needed correcting

- `selected` **is** set to `true` on backend proposals (`CommandProposal.selected` defaults `True`), so the fix must not lean on it being absent. There's a test for the `selected: true` variant.
- The `hideRejected` toggle **does** render for an assess-only dismissal — `summary.js:3510` already includes `'assess'`. The documented provider workaround is genuinely reachable.
- `background` is **not** server-validated (`AssessParser.validate` checks only `narrative`; SDK `AssessCommand.background` has no `max_length`), so only `narrative` matters for the length issue.

## Testing

- `tests/static/command-drop.test.mjs` — 11 new tests in the only CI-covered JS location (`node --test 'tests/static/**/*.test.mjs'`), using the stuck record's exact cached shape as the regression case. **28/28 pass.**
- `test_session_view.py` — structural pin matching this file's existing `summary.js` pin convention, including a negative assertion against the old inline ternary. **209/209 pass.**
- `./lint.sh` (ruff check, ruff format, mypy, pytest) green.

**Not verified:** no live repro. The ticket's own "Steps to reproduce" section is unfilled, and its mechanism was derived from deployed source rather than observed. Everything here comes from source, the SDK models, and the two test suites — a manual pass is still worth doing (see below).

## Release notes

Purely client-side. `ScribeSummary.commands` is an opaque `JSONField` round-tripped verbatim and `NoteLockGuard` only reads `approved`, so **no migration or backfill**.

But it is **not self-healing**: `approved` only flips via the client POST, so each of the 11 stuck notes needs someone to open the Scribe tab and click Approve once. Static JS is served with no cache-busting or `Cache-Control`, so that pass needs a **hard reload**.

Manifest bumped to `2026-07-29 v0.4.43`. Also repairs the `tags` block, which had drifted (`2026-07-27` / `0.4.41` while `plugin_version` already read `v0.4.42`).

## Manual verification still to do

- [ ] Dismiss a duplicate condition → Approve succeeds
- [ ] Dismissed condition is absent from the chart after insert
- [ ] Restore-by-click still works, and the Hide Dismissed toggle still appears
- [ ] A genuinely invalid command (e.g. incomplete Rx) still blocks Approve with its specific message

## Follow-ups worth filing separately

- Condition cards have **no delete**, only dismiss (`renderConditionActions` accepts only `onDismiss`).
- A non-dismissed assess with `narrative > 2048` has no client gate and 400s the whole insert batch server-side.
- `AssessParser.build` passes `condition_id=data.get("condition_id") or ""` with no presence check.
- `dismissAssess` leaves `accepted: true` beside `rejected: true`. Currently harmless — all consumers check `rejected` first — but contradictory, and it touches 3+ read sites, so it was left out of a P0.

[KOALA-6555]: https://canvasmedical.atlassian.net/browse/KOALA-6555?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/hyperscribe/CANVAS_MANIFEST.json
+++ b/hyperscribe/CANVAS_MANIFEST.json
@@ -1,6 +1,6 @@
 {
   "sdk_version": "0.164.0",
-  "plugin_version": "2026-07-28 v0.4.42",
+  "plugin_version": "2026-07-29 v0.4.43",
   "name": "hyperscribe",
   "description": "Create commands based on the content of an audio discussion between a patient and a provider.",
   "url_permissions": [
@@ -322,9 +322,9 @@
     "VisitTemplates"
   ],
   "tags": {
-    "version_date": "2026-07-27",
-    "version_branch": "koala-5808",
-    "version_semantic": "0.4.41"
+    "version_date": "2026-07-29",
+    "version_branch": "koala-6555",
+    "version_semantic": "0.4.43"
   },
   "custom_data": {
     "namespace": "canvas_medical__scribe",
--- a/hyperscribe/scribe/static/command-drop.js
+++ b/hyperscribe/scribe/static/command-drop.js
@@ -1,0 +1,51 @@
+// Why a command was left out of the /insert-commands batch — the single source of
+// truth for "is this drop the user's choice, or a silent failure we must surface?"
+//
+// handleInsert splits `commands` into `insertable` and `dropped`. Some drops are
+// deliberate (the user unchecked the row, the row is empty, the user dismissed the
+// card); one is not (the row failed validation and would be lost without a word).
+// Only the last kind may block Approve.
+//
+// KOALA-6555: this module exists because that distinction used to be hand-written in
+// two places — the COMMANDS_FILTERED audit reason and the `droppedForValidation`
+// filter — and they drifted. When the ✕ on a condition card started marking
+// `data.rejected` instead of deleting the row, a new *intentional* drop reason was
+// created; neither copy was taught about it, so dismissing a duplicate condition
+// classified as a validation failure and made the note permanently unsignable.
+// Both call sites now read from here, so a new drop reason cannot be added to one
+// without the other.
+//
+// Zero imports on purpose (same as bmi.js / questionnaire-score.js): summary.js is
+// not importable under Node, so this is the only way the classification can carry
+// real unit tests (tests/static/command-drop.test.mjs).
+
+// The two condition-card families. Both render a ✕ that marks `data.rejected` rather
+// than deleting the row, so the assessment text survives a mis-click and the card can
+// be restored (see renderConditionActions in soap-group.js).
+export const CONDITION_COMMAND_TYPES = new Set(['assess', 'diagnose']);
+
+// True when the provider dismissed a condition card with the ✕. Deliberately scoped to
+// the condition families rather than testing `data.rejected` on any type: suppressing a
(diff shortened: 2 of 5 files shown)


kevincareycanvas:
cerberus review

PR #335, opened 2026-07-29
feat(scribe): remove the Hyperscribe Customization app

kevincareycanvas:
## Summary

Removes the **Hyperscribe Customization** app — the global app-drawer entry that rendered a panel into the chart's right pane, letting users tune per-command LLM prompts and pick a default UI tab for the legacy Hyperscribe capture view. None of it applies to the Canvas Scribe experience, so it's deprecated on this branch.

There was only ever *one* app here: `CustomizationApp` was registered with `scope: global` / `show_in_panel: false` and launched into `RIGHT_CHART_PANE`, so the drawer entry and the chart-side panel were the same component.

`hyperscribe/scribe/**` had zero references to `Customization` or `custom_prompts` — the only non-app consumer was the legacy `handlers/capture_view.py`.

## What was removed

**The app surface**

- `handlers/customization_app.py` — the `Application`
- `handlers/customization_display.py` — its SimpleAPI (`GET /customization/all`, `POST /customization/command`, `POST /customization/ui_default_tab`)
- `templates/customization.html`, `static/customization.png`

**The S3 read path it fed** — dead code once nothing can write customizations

- `libraries/customization.py` — read/write of `hyperscribe-{instance}/customizations/customizations_{user}.json`
- `structures/customization.py`, `structures/default_tab.py`

**Edits**

- `CANVAS_MANIFEST.json` — dropped the `CustomizationDisplay` protocol (22 remain), collapsed `applications` to `[]`, bumped to `v0.4.44`
- `handlers/capture_view.py` — removed the `Customization.customizations()` lookup, the `uiDefaultTab` context key, and the `custom_prompts_as_secret` merge in `run_commander`
- `templates/hyperscribe.html` — Transcript hardcoded active

## Deliberately kept

Per-command custom prompts stay configurable **instance-wide via the `CustomPrompts` plugin secret**. Only the per-user UI and its S3 storage go away. So `structures/custom_prompt.py`, `Settings.custom_prompts`, `Constants.SECRET_CUSTOM_PROMPTS`, `commands/base.py`'s `custom_prompt()` / `command_from_json_custom_prompted()`, the `command_custom_prompt` JSON schema, and the five commands that use them (`follow_up`, `history_of_present_illness`, `instruct`, `plan`, `reason_for_visit`) are all untouched.

## Behavior changes to be aware of

1. **`CustomPrompts` is now the single source of truth.** In `run_commander` the per-user S3 blob previously *overrode* the secret. Any instance relying on per-user S3 prompts now falls back to the instance-wide secret value.
2. **Default tab is fixed to Transcript** in the legacy capture view. That was already the code default (`DefaultTab.TRANSCRIPT`), so it's a no-op for anyone who never opened the app.
3. **No S3 cleanup.** Existing `hyperscribe-{instance}/customizations/*.json` objects are left in place and simply stop being read — worth a follow-up if you want them purged.

## Testing

- `pytest tests/` → **2880 passed**
- `./lint.sh` (ruff check, ruff format, mypy, pytest) → all passed
- Manifest parses as valid JSON; a final `grep -i customiz` sweep leaves only the unrelated Nabla `note_sections_customization` code and the `VisitTemplates` README prose

Note: bare `pytest` (no path) fails at collection with `KeyError: 'APISigningKey'` — that's pre-existing on `feat/canvas-scribe` and needs `local_env.sh` sourced; verified against a clean tree.


kevincareycanvas:
cerberus review

kevincareycanvas:
janus uat

kevincareycanvas:
cerberus review

PR #336, opened 2026-07-30
feat(scribe): polish the finalized read-only note view (KOALA-5631)

kevincareycanvas:
## Problem

Once a note is finalized, the Scribe view reads as *disabled* rather than *done* — every card at 60% opacity, a padlock in each corner, an "Already in chart ✓" badge on each row.

That treatment belongs to **amend mode**, where the dim distinguishes rows already on the note from rows you can still edit. The predicate behind it (`rowReadOnly && already_documented`) happens to be true for *every* card once `approved === true`, so it applies to everything and communicates nothing. Same for the settled badge: read-only mode already filters to only inserted items (`soap-group.js:776-781`), so "Already in chart" describes 100% of what's on screen.

## Approach

Gate the chrome behind a new `.summary-container--finalized` class rather than rewriting the 30+ sites in `soap-group.js` that emit it.

Those sites are correct while amending — only the finished-document *presentation* changes. Scoping every rule under a class that exists solely when `approved === true` makes it structurally impossible to affect amend mode or the locked-but-incomplete state, and no predicate governing editability (`rowLocked`, `readOnly`, `<fieldset disabled>`, server-side `_authorize_edit`) is touched. The trade-off is dead DOM: the padlock SVG and badge still render, hidden. Removing those nodes is a clean follow-up against a design that's already settled.

`approved` comes from the summary payload rather than the viewer, so the author's finalized view and a colleague's read-only view of the same note both get the treatment from one condition.

## Changes

**`styles.css`** — one `.summary-container--finalized` block: full contrast, no padlock, no settled badge, no hover backgrounds, no pointer cursors on cards that can't open, and the charges-table editing residue (grips, ghost checks, lock column) cleared.

Medication and allergy rows needed a second rule. Their view branches (`medication-row.js:145`, `allergy-row.js:146`) add a `documented` class carrying an **independent** 0.55 opacity that stacks with the card-level dim — so clearing only `.command-locked` left those two card types grey while everything around them went full contrast.

**`summary.js`** — three small edits: the container class, transcript collapsed by default on a note that *arrives* finalized (set in both the `initialData` and cache-restore paths, so it can't fire when someone signs in-session and yank the panel shut), and the success verification banner no longer reappearing on revisit. It still shows at the moment of signing, and failures still surface in both paths. That change removed the effect's run-once latch, so it gets an explicit `autoVerifiedRef` — otherwise it would re-POST `/verify-commands` every time `syncNoteCommands` returns fresh arrays.

**`soap-group.js`** — generalize the raw-section-text guard from covered *history* keys to any key a review command covers. `HISTORY_SECTION_KEYS` omits `current_medications`, `allergies`, and `social_history`, so each rendered a second, unstyled copy below its card whenever the dedicated branch found nothing to show. Same reasoning as the existing `assessment_and_plan` exclusion directly above it: the card is the live, editable copy; the raw text is a stale duplicate.

## Reviewer notes

- **The `soap-group.js` change is the one to scrutinize.** It's the only change that also affects the **editing** view, and it has no test coverage — the logic sits inside a large Preact render. Happy to add a structural pin in `test_session_view.py` alongside the existing `summary.js` marker pins if that's wanted.
- Nothing editable is lost by the suppression: the removed copy was a `<p>`; the live copy is the Chart Review / History Review card.
- `current_medications` and `allergies` never reach that fallback while editing — `onAddMedication` / `onAddAllergy` are non-null there and satisfy the guard at `soap-group.js:1287`.

## Verification

`./lint.sh` passes (ruff, ruff format, mypy, 2900 pytest). `node --test 'tests/static/**/*.test.mjs'` passes 28.

Reviewed against a real finalized note during development, which is how the `.documented` double-dim surfaced.

**Not yet exercised in a running instance** — worth covering before merge:
- [ ] Amend mode: click "Make changes" and confirm padlocks and dimming **return** on already-documented rows
- [ ] Note locked with Scribe incomplete: container dim and "click Amend to finish" banner unchanged
- [ ] The duplicate-suppression fix on a note with `_chart_review` / `_history_review` content
- [ ] Transcript collapsed on load; success banner present at signing, absent on reload

### DIFF
--- a/hyperscribe/scribe/static/soap-group.js
+++ b/hyperscribe/scribe/static/soap-group.js
@@ -819,7 +819,6 @@ export function SoapGroup({ title, groupColor, sections, commandBySectionKey, on
              (key === 'prescription' && visibleRecs.some(r => r.command_type === 'prescribe')));
           const DEDICATED_SECTION_KEYS = new Set(['current_medications', 'allergies']);
           const HISTORY_SECTION_KEYS = new Set(['past_medical_history', 'past_surgical_history', 'family_history']);
-          const isCoveredHistory = coveredKeys.has(key) && HISTORY_SECTION_KEYS.has(key);
           if (coveredKeys.has(key) && !hasRecsForKey && !DEDICATED_SECTION_KEYS.has(key) && !HISTORY_SECTION_KEYS.has(key)) return null;
           const cmds = commandBySectionKey && commandBySectionKey[key];
 
@@ -1509,7 +1508,16 @@ export function SoapGroup({ title, groupColor, sections, commandBySectionKey, on
           // plan commands), so raw `s.text` is always a duplicate — and it becomes a stale,
           // uneditable duplicate once every card is moved to Wrap Up (which empties
           // assessment_and_plan of commands and drops rendering into this fallback).
-          const showHistoryText = s.text && !isCoveredHistory && key !== 'assessment_and_plan';
+          // Never print raw note text for a key a review command already renders
+          // inside its card (chart_review covers allergies / current_medications /
+          // immunizations; history_review covers the four history keys; see
+          // getCoveredKeys). Same reasoning as the assessment_and_plan exclusion
+          // above: the card is the live, editable copy, so the raw section text is
+          // a stale uneditable duplicate. This used to check covered HISTORY keys
+          // only, which left current_medications, allergies, and social_history
+          // printing a second, unstyled copy below their card whenever the
+          // dedicated branch above found nothing to render (KOALA-5631 follow-up).
+          const showHistoryText = s.text && !coveredKeys.has(key) && key !== 'assessment_and_plan';
           // social_history has no manual input affordance (not in NARRATIVE_SECTIONS,
           // no SECTION_TO_HISTORY_TYPE entry), so an empty Social History section is a
           // dead, un-fillable header. Hide it whenever it has nothing to show; the AI
--- a/hyperscribe/scribe/static/styles.css
+++ b/hyperscribe/scribe/static/styles.css
@@ -426,6 +426,92 @@ h2 { margin-bottom: 4px; }
   cursor: not-allowed;
 }
 
+/* Finalized notes (approved === true). Every card is on the note and nothing is
+   editable, so amend mode's per-card dim + padlock + "Already in chart" badge
+   describe 100% of the view and inform nothing — a signed note ends up looking
+   like a form someone greyed out. soap-group.js still EMITS all of that chrome
+   (it's correct while amending, where the dim contrasts locked rows against
+   still-editable ones — see rowLocked/ICON_LOCK in soap-group.js); this block
+   suppresses it for the finished document. Everything here is scoped to
+   --finalized, which only exists when approved, so amend mode and the
+   locked-but-incomplete state (.summary-container--readonly above) are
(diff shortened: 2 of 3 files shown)


kevincareycanvas:
cerberus review

kevincareycanvas:
janus uat

kevincareycanvas:
cerberus review

kevincareycanvas:
janus uat

kevincareycanvas:
cerberus review

kevincareycanvas:
cerberus review

kevincareycanvas:
Bottom Line: all six UAT checks pass on `scribeqa-playground` at `hyperscribe@2026-08-25 v0.4.49`. One pre-existing gap surfaced during testing that is worth knowing about, but it does not block this PR.

Tested against a signed note with inserted Scribe commands, charges, and Chart Review content.

## Wins

- The auto-verify retry fix is confirmed working. With `/verify-commands` blocked in devtools, three attempts fired from `summary.js:1416` inside a single page session: blocked, blocked, then a 200 once I unblocked it. `DOMContentLoaded` and `Load` held steady across both observations, so no reload accounts for it, and on the pre-fix build the latch would have allowed exactly one attempt with nothing after.
- Amend mode still dims correctly. Clicking "Make changes" brings the padlocks and the dimming back on rows already in the chart, which is what the CSS gating was supposed to preserve.
- A note locked with Scribe incomplete is unchanged, with the dim and the "click Amend to finish" banner both intact. That is the check that proves the new `--finalized` class did not leak into a state it should never reach.
- No duplicate section text appears under the Chart Review or History Review cards. Med List Updates and Allergies are the two that actually changed behavior, and both are clean.
- Charges rank numbers now read at the same weight as the diagnosis labels beside them. Previously every row was locked, so every number sat faded next to a label at full contrast.
- The transcript arrives collapsed with its entry count visible, and the success banner is absent on reload. Both halves of that behave as intended.

## Issues Identified

Transparently, I could not verify that the success banner appears at the moment of signing, because it never appears at all.

`handleInsert` sets the banner state at `summary.js:3094`, awaits `/configure-command-buttons` at 3140, then posts `CLOSE_MODAL` at 3153. The comment already sitting in the code at 3137 says it directly: once `CLOSE_MODAL` lands, this component is unmounted. So the banner renders for the length of one fetch and then the component goes away, which is why nobody sees it.

The failure banner is set on the same line and shares that fate. A command that silently failed to insert is therefore invisible to the provider at the moment they sign. `logEvent('COMMANDS_FAILED')` at 3104 still records it, so we can detect it after the fact, but the clinician gets no signal in the moment. Failures do surface on the next visit to the note through the load-time auto-verify.

None of this is new on this branch. `CLOSE_MODAL` predates it, and the latch fix in `c20201be` actually improves the deferred path, because before it a single transient failure on that load-time verify meant the failure never surfaced at all.

## Key Learning

A real fix has to move the notification outside the component's lifetime rather than delay the close, since the unmount is deliberate and documented. Holding the close when `failedCount > 0`, or surfacing failures somewhere outside the Scribe iframe, are the two directions I can see. Both reach into the home-app side, so I did not scope either one into this PR.

## Next Steps

- Kevin Carey: merge into `feat/canvas-scribe` once review comes back clean.
- Open question for reviewers: do we want the success banner fixed, or deleted? It shows nowhere today and nobody has reported missing it, so I lean toward dropping it and putting the effort into the failure path instead. Tell me if there is a better pattern.

PR #340, opened 2026-08-12
fix(scribe): recover dictated PRN medications from the transcript (KOALA-6644)

ruipbatista:
https://canvasmedical.atlassian.net/browse/KOALA-6644

## The bug

Canvas Scribe charted medications from the **Nabla-generated note**, never from the
**transcript**. When Nabla's summarization omitted a medication it could never be charted, and
nothing raised an error to the provider, the logs, or the audit trail — the only signal was
absence. Loss rate rises with the number of PRNs dictated; the ticket measured 512 Brigade
notes losing PRN content, 24% at 3+ PRNs.

Both medication recommenders already *received* a `transcript` argument and ignored it.

## What changed

**Recovery (`recommendations/medication_statement.py`)** — scans the transcript for as-needed
phrasing and sends the matching excerpts alongside the note sections, following the windowing
pattern `TaskRecommender` already established. New `is_prn` / `from_transcript` schema fields;
prompt gains completeness pressure, a PRN instruction, and an explicit *same drug, two orders*
rule so a scheduled order can never stand in for an as-needed one.

**Scope: medication list only.** The prescription path deliberately does **not** recover PRNs.
Nabla writes continuation as `Continue <drug> as needed ...`, so the "Continue" marker sits on
the very line that gets dropped — a recovered PRN arrives with no evidence of whether it is
being started or merely reconciled. Measured on real notes, that path proposed new
prescriptions for five medications the patient was already taking. Missing documentation is
recoverable; a duplicate prescription is not, and the ticket's confirmed case was a
medication-list loss. `prescription.py` carries a comment recording this so it is not re-added.

**Dedup fix (`commands/medication_statement.py`)** — `annotate_duplicates` stamped
`already_documented` on a drug-name substring match against the chart, and the UI hides
`already_documented && !command_uuid`. So recovering a PRN was a **no-op** whenever the same
drug was already charted on a schedule, which is exactly the reported case. The chart sig is now
read in the same query and an as-needed proposal is suppressed only on positive evidence that
the chart already carries an as-needed order. Non-PRN proposals dedupe on name exactly as before.

**Provenance is established, not asserted.** Two deterministic corrections, both found by
measuring against real notes rather than by reasoning:

- The model sets `fromTranscript` even when no excerpts were supplied at all, which would badge a
  medication as recovered on a note where no transcript was read.
- On **partial** loss it claims recovery even when the note still carries the order. Measured: Nabla
  dropped all six dictated PRNs from `CURRENT_MEDICATIONS` but kept every one in
  `ASSESSMENT_AND_PLAN`, which this recommender also reads — and all six came back flagged, with the
  audit reporting 6 recoveries where the honest answer was 0. The ticket splits its 512 notes into
  252 dropped entirely and 260 dropped partially, so this is the common case, not an edge.

`note_documents_as_needed()` clears the flag when a supplied section already documents the drug
**with as-needed dosing**. Name *and* as-needed, never name alone: a drug routinely appears in the
note under a scheduled order while its as-needed order is the one that went missing — the reported
lorazepam case — so a name-only test would kill genuine recoveries.

**Telemetry** — `MEDS_PRN_TRANSCRIPT` records as-needed mentions against actual recoveries.
Mentions with zero recoveries is the remaining-silent-drop case and is what a later remediation
sweep can select on. Emitted *outside* the recommendation `try`, which swallows the whole stage
as "non-critical" — inside it, the telemetry would vanish in exactly the runs where a medication
went missing. Counts only: no drug names, no sigs, no transcript text.

Plus a UI pill on recovered rows, and `post_recommend_commands` now supplies the transcript
(without it the standalone endpoint reintroduced the bug).

First commit is a pure refactor lifting the windowing helpers out of `task.py`; `test_task.py`
passes unedited as the proof.

## Validation

Measured against real notes, comparing the same stored note and transcript with the PRN lines
stripped from the note to simulate Nabla's loss. Both directions of the provenance guard:

| condition | med-list proposals | flagged recovered |
|---|---|---|
full note, no transcript | 5 | 0 |
full note, **with** transcript | 10 | **0** — was 6 false positives before the guard |
stripped note, no transcript | 8, several with an empty sig | 0 |
stripped note, **with** transcript | 10 | **5** genuine recoveries, still flagged |

Row 2 is the guard working: the transcript still improves med-list completeness (5 → 10 entries,
because the note filed its PRNs under prescriptions rather than the medication list) while
correctly reporting none of them as recovered, since the note did document them. Row 4 is the guard
not over-reaching: with the PRN lines genuinely gone, recovery fires and stays flagged.

Row 3 is the pre-fix failure in its subtler form — the drug names survive elsewhere in the note but
their directions do not, so the medication list carries entries with no sig at all.

Across all runs the two lorazepam orders stay separate (scheduled unflagged, as-needed flagged) and
decoy phrasing — "we'll follow up as needed", "call the office if needed" — produces no medication.
With no as-needed phrasing anywhere the prompts are byte-identical to before, which the pre-existing
prompt assertions still pin.

**Caveat worth carrying into review:** across three live dictation/fixture runs, Nabla never once
dropped a PRN from *every* section this recommender reads — it kept them in the assessment/plan each
time. Recovery has been exercised only via the controlled strip test. This is a backstop for a
failure mode reproduced from the ticket's data, not one we have caught end-to-end ourselves.

## Not in scope

- **The affected notes are not repaired.** Fix-forward only. A remediation sweep is agreed as
  follow-up work once this is validated in the wild, and there is currently no host for a batch
  job in this plugin.
- **The Nabla-side root cause.** `note_sections_customization` has no entry for
  `CURRENT_MEDICATIONS`, so meds are generated with stock template behaviour. Worth a spike.
- **KOALA-5411** (vitals parser) — same symptom, different mechanism: that data *is* in the note
  and a regex misses it.

## Pre-existing issues found while testing — please don't block on these

Both reproduce on the base branch and need their own tickets:

- A lorazepam order dictated as "one tablet daily" with no strength resolves to **Ativan 1 mg** —
  the resolver falls through to the first FDB candidate. KOALA-5930 only made this fail-safe when
  a strength *was* stated and didn't match. Wrong strength on a benzodiazepine.
- **"Acetaminophen 650 mg" resolves to "Dologen 650 mg-2 mg"**, a combination product rather than
  plain acetaminophen: 650 mg matches the combination's component so it passes the strength check
  while being the wrong drug. Arguably higher severity than this ticket — wrong-drug beats
  missing-drug.

## UAT

**A happy-path run will not test this.** Nabla keeps the PRNs roughly 3 runs in 4 — it kept all
six in both local dictation and fixture runs — and then there is nothing to recover, everything
is correctly unflagged, and the run passes green without touching the change. Testing recovery
requires forcing the note to omit the PRNs (edit `summary.note_data` via the Models tab / the
`debug-cache` endpoint, then re-run `/recommend-commands`). A scenario-by-scenario UAT script
with pass/fail criteria is available — ask and I'll paste it into a comment.

### DIFF
--- a/hyperscribe/scribe/api/session_view.py
+++ b/hyperscribe/scribe/api/session_view.py
@@ -47,6 +47,7 @@ from hyperscribe.scribe.clients.nabla.backend import NablaBackend
 from hyperscribe.scribe.backend import (
     ClinicalNote,
     CodingEntry,
+    CommandProposal,
     Condition,
     NoteSection,
     Observation,
@@ -86,6 +87,7 @@ from hyperscribe.scribe.commands.problem_list_match import (
 )
 from hyperscribe.scribe.recommendations import make_llm_client, prescription_dispense_enabled, recommend_commands
 from hyperscribe.scribe.recommendations._referral_diagnosis import link_referral_diagnoses
+from hyperscribe.scribe.recommendations._transcript_windows import PRN_PATTERN, find_keyword_matches
 from hyperscribe.scribe.recommendations.diagnosis_llm_resolver import (
     BlockContext as DiagnosisBlockContext,
     resolve_uncoded_blocks,
@@ -1368,6 +1370,7 @@ class ScribeSessionView(StaffSessionAuthMixin, SimpleAPI):
         # ── Step 3: Recommend commands ──
         _save_progress(note_id, 3, total, SUMMARY_STEPS[3])
         recommendations_list: list[dict[str, Any]] = []
+        rec_proposals: list[CommandProposal] = []
         api_key = self.secrets.get("AnthropicAPIKey", "")
         if api_key:
             try:
@@ -1403,6 +1406,29 @@ class ScribeSessionView(StaffSessionAuthMixin, SimpleAPI):
                 ]
             except Exception:
                 log.exception("recommend_commands failed (non-critical)")
+
+        # KOALA-6644: the defect was invisible — a dropped PRN produced no error, no log
+        # line and no audit row, so the only signal was absence. Record how many as-needed
+        # mentions the transcript carried against how many medications were actually
+        # recovered from it. Mentions with zero recoveries is the remaining-silent-drop
+        # case, and is what a later remediation sweep can select on.
+        #
+        # Deliberately outside the try above: that block swallows the entire recommendation
+        # stage as "non-critical", so emitting inside it would lose the telemetry in exactly
+        # the runs where a medication went missing. Emitted only when the transcript
+        # mentioned as-needed dosing at all, so the audit log stays signal-dense (mirrors
+        # VITALS_FIELD_REFUSED). Counts only — no drug names, no sigs, no transcript text.
+        if note_uuid and transcript:
+            prn_mentions = len(find_keyword_matches(transcript, PRN_PATTERN))
+            if prn_mentions:
+                audit_event(
+                    note_uuid,
+                    "MEDS_PRN_TRANSCRIPT",
+                    {
+                        "prn_mentions": prn_mentions,
+                        "recovered_from_transcript": sum(1 for p in rec_proposals if p.from_transcript),
+                    },
+                )
 
         # ── Step 3b: Check medication interactions ──
         interaction_warnings: list[dict[str, Any]] = []
--- a/hyperscribe/scribe/api/session_view.py
+++ b/hyperscribe/scribe/api/session_view.py
@@ -1307,6 +1307,7 @@ class ScribeSessionView(StaffSessionAuthMixin, SimpleAPI):
                 "selected": p.selected,
                 "section_key": p.section_key,
                 "already_documented": p.already_documented,
+                "from_transcript": p.from_transcript,
             }
             for p in proposals
         ]
@@ -1396,6 +1397,7 @@ class ScribeSessionView(StaffSessionAuthMixin, SimpleAPI):
                         "selected": p.selected,
                         "section_key": p.section_key,
                         "already_documented": p.already_documented,
+                        "from_transcript": p.from_transcript,
                     }
                     for p in rec_proposals
                 ]
@@ -1641,6 +1643,7 @@ class ScribeSessionView(StaffSessionAuthMixin, SimpleAPI):
                             "selected": p.selected,
                             "section_key": p.section_key,
                             "already_documented": p.already_documented,
+                            "from_transcript": p.from_transcript,
                         }
                         for p in proposals
                     ],
@@ -1669,9 +1672,26 @@ class ScribeSessionView(StaffSessionAuthMixin, SimpleAPI):
         zip_codes = resolve_zip_codes(patient_id, rec_note_id) or None
         allowlist = self.secrets.get(Constants.SECRET_SCRIBE_PRESCRIPTION_STAFFERS, "")
         dispense_engine_enabled = prescription_dispense_enabled(allowlist, _note_provider_id(rec_note_id))
+        # The medication recommenders fall back to the transcript when the generated note has
+        # dropped a PRN, so this endpoint must supply it too — otherwise re-running
+        # recommendations standalone silently reintroduces KOALA-6644.
+        transcript_data = data.get("transcript", {})
+        if not transcript_data.get("items") and rec_note_id:
+            try:
+                transcript_data = _load_transcript(rec_note_id)
+            except Exception:
+                # Recovering PRNs is an enhancement over the note-only path; never let a
+                # missing or unreadable transcript fail the whole endpoint.
+                log.exception("Could not load transcript for recommendations; falling back to note only")
+                transcript_data = {}
+        transcript = _parse_transcript(transcript_data) if transcript_data.get("items") else None
         try:
             proposals = recommend_commands(
-                note, api_key, zip_codes=zip_codes, dispense_engine_enabled=dispense_engine_enabled
+                note,
+                api_key,
+                zip_codes=zip_codes,
+                transcript=transcript,
+                dispense_engine_enabled=dispense_engine_enabled,
             )
         except Exception:
             log.exception("recommend_commands failed")
@@ -1695,6 +1715,7 @@ class ScribeSessionView(StaffSessionAuthMixin, SimpleAPI):
                             "selected": p.selected,
                             "section_key": p.section_key,
                             "already_documented": p.already_documented,
+                            "from_transcript": p.from_transcript,
                         }
                         for p in proposals
                     ],
--- a/hyperscribe/scribe/backend/models.py
+++ b/hyperscribe/scribe/backend/models.py
@@ -110,6 +110,7 @@ class CommandProposal:
         selected: bool = True,
         section_key: str = "",
         already_documented: bool = False,
+        from_transcript: bool = False,
     ) -> None:
         self.command_type = command_type
         self.display = display
@@ -117,6 +118,9 @@ class CommandProposal:
         self.selected = selected
         self.section_key = section_key
         self.already_documented = already_documented
+        # True when this proposal was recovered from the visit transcript because the
+        # generated note omitted it. Drives the "from transcript" badge in the review UI.
+        self.from_transcript = from_transcript
 
 
 class PatientContext:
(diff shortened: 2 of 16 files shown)


ruipbatista:
janus uat verify Scribe was silently dropping dictated as-needed (PRN) medications: it extracted meds only from
the Nabla-generated note, so anything Nabla's summary omitted was never charted and nothing
errored. The fix recovers them from the transcript into the medication list.

IMPORTANT: a normal scribe run will NOT test this. Nabla keeps the PRNs about 3 runs in 4, and
then there is nothing to recover — everything is correctly unflagged and the run passes green
without exercising the fix. You must force the note to omit the PRNs (step 3).

Steps
1. On a test note, set the transcript to the attached fixture with finalized: true — Models tab,
   or PUT /plugin-io/api/hyperscribe/scribe-session/debug-cache
   {"note_id":"<uuid>","type":"transcript","fields":{"items":[...],"finalized":true}}
   Needs the test staff key in ScribeDebugStaffers. finalized must be true or Generate rejects it.
2. Generate.
3. In the Models tab edit summary.note_data: delete every line naming a PRN drug together with
   as-needed dosing, across all sections. KEEP the scheduled lorazepam line ("one tablet daily
   ... before showers") — it has no as-needed wording and must survive.
4. Re-run POST /plugin-io/api/hyperscribe/scribe-session/recommend-commands.

Expect
- The six PRNs reappear in the medication list with from_transcript: true and a "From transcript"
  pill, even though the note no longer contains them.
- Two separate lorazepam rows: scheduled (no pill) and as-needed (pill). Merged into one = FAIL.
- No recovered PRN appears as a NEW PRESCRIPTION. Recovery is medication-list only; a recovered
  PRN proposed as a prescription = FAIL.
- No medication proposed from the decoys "we'll follow up as needed" / "call the office if needed".
- Audit tab shows MEDS_PRN_TRANSCRIPT with counts only. Any drug name or sig in that payload is a
  PHI failure.

Do not fail the PR for these (pre-existing on base)
- Lorazepam's shower dose resolving to "Ativan 1 mg" when no strength was dictated.
- "Acetaminophen 650 mg" resolving to "Dologen 650 mg-2 mg" (a combination product).

If you can also add an active lorazepam with a SCHEDULED sig to the patient chart before step 3,
do it — the recovered as-needed order must still be visible and NOT marked "Already in chart".
That dedup path is what made recovery a no-op before the fix.

ruipbatista:
janus uat verify Scribe silently dropped dictated as-needed (PRN) medications: it extracted meds only from the
Nabla-generated note, so anything the summary omitted was never charted and nothing errored. The
fix reads the transcript as a backstop and recovers them into the medication list, badged
"From transcript".

IMPORTANT: a normal scribe run will NOT test this. Nabla usually keeps the PRNs somewhere in the
note, and then there is nothing to recover — everything is correctly unbadged and the run passes
green without exercising the fix. You must edit the generated note yourself. Two tests below,
same setup, differing only in how much you delete.

Setup (once)
1. On a test note set the transcript to the attached fixture with finalized: true — Models tab, or
   PUT /plugin-io/api/hyperscribe/scribe-session/debug-cache
   {"note_id":"<uuid>","type":"transcript","fields":{"items":[...],"finalized":true}}
   Needs the test staff key in ScribeDebugStaffers. finalized must be true or Generate rejects it.
2. Generate. Note which sections of summary.note_data contain the six PRNs.

TEST A — partial drop must NOT produce false badges (the common case)
3. In the Models tab, delete the PRN lines from CURRENT_MEDICATIONS only. Leave
   ASSESSMENT_AND_PLAN and PRESCRIPTION untouched.
4. Re-run POST /plugin-io/api/hyperscribe/scribe-session/recommend-commands.
EXPECT: the PRNs are present in the medication list, and NONE of them shows a "From transcript"
pill. Audit MEDS_PRN_TRANSCRIPT shows recovered_from_transcript: 0.
FAIL: any pill, or a non-zero recovered count. The note still documented those orders, so claiming
recovery is false provenance. This was observed on a real note before the fix — six false pills and
a count of 6 where the truth was 0.

TEST B — full drop must recover (the fix itself)
5. Same note. Now also delete every PRN line from ASSESSMENT_AND_PLAN and PRESCRIPTION. KEEP the
   scheduled lorazepam line ("one tablet daily ... before showers") — it has no as-needed wording
   and must survive.
6. Re-run recommend-commands.
EXPECT: the PRNs reappear in the medication list WITH "From transcript" pills and their directions
intact; two separate lorazepam rows — scheduled (no pill) and as-needed (pill); a non-zero
recovered_from_transcript.
FAIL: PRNs stay missing; or they come back with no sig; or a recovered row has no pill; or the two
lorazepam orders merge into one.

Applies to both
- No recovered PRN may appear as a NEW PRESCRIPTION. Recovery is medication-list only by design.
- No medication from the decoys "we'll follow up as needed" / "call the office if needed".
- MEDS_PRN_TRANSCRIPT must contain counts only. Any drug name or sig in that payload is a PHI fail.

Do not fail the PR for these (all reproduce on the base branch)
- "Acetaminophen 650 mg" resolving to "Dologen 650 mg-2 mg", a combination product.
- Lorazepam's shower dose resolving to "Ativan 1 mg" when no strength was dictated.
- The same PRN appearing as both a medication-list entry and a prescription when Nabla's own note
  lists it in both places — check from_transcript to tell that apart from a scope regression.

ruipbatista:
janus uat verify Scribe silently dropped dictated as-needed (PRN) medications: it extracted meds only from the
Nabla-generated note, so anything the summary omitted was never charted and nothing errored. The
fix reads the transcript as a backstop and recovers them into the medication list, badged
"From transcript".

IMPORTANT: a normal scribe run will NOT test this. Nabla usually keeps the PRNs somewhere in the
note, and then there is nothing to recover — everything is correctly unbadged and the run passes
green without exercising the fix. You must edit the generated note yourself. Two tests below,
same setup, differing only in how much you delete.

Setup (once)
1. On a test note set the transcript to the attached fixture with finalized: true — Models tab, or
   PUT /plugin-io/api/hyperscribe/scribe-session/debug-cache
   {"note_id":"<uuid>","type":"transcript","fields":{"items":[...],"finalized":true}}
   Needs the test staff key in ScribeDebugStaffers. finalized must be true or Generate rejects it.
2. Generate. Note which sections of summary.note_data contain the six PRNs.

TEST A — partial drop must NOT produce false badges (the common case)
3. In the Models tab, delete the PRN lines from CURRENT_MEDICATIONS only. Leave
   ASSESSMENT_AND_PLAN and PRESCRIPTION untouched.
4. Re-run POST /plugin-io/api/hyperscribe/scribe-session/recommend-commands.
EXPECT: the PRNs are present in the medication list, and NONE of them shows a "From transcript"
pill. Audit MEDS_PRN_TRANSCRIPT shows recovered_from_transcript: 0.
FAIL: any pill, or a non-zero recovered count. The note still documented those orders, so claiming
recovery is false provenance. This was observed on a real note before the fix — six false pills and
a count of 6 where the truth was 0.

TEST B — full drop must recover (the fix itself)
5. Same note. Now also delete every PRN line from ASSESSMENT_AND_PLAN and PRESCRIPTION. KEEP the
   scheduled lorazepam line ("one tablet daily ... before showers") — it has no as-needed wording
   and must survive.
6. Re-run recommend-commands.
EXPECT: the PRNs reappear in the medication list WITH "From transcript" pills and their directions
intact; two separate lorazepam rows — scheduled (no pill) and as-needed (pill); a non-zero
recovered_from_transcript.
FAIL: PRNs stay missing; or they come back with no sig; or a recovered row has no pill; or the two
lorazepam orders merge into one.

Applies to both
- No recovered PRN may appear as a NEW PRESCRIPTION. Recovery is medication-list only by design.
- No medication from the decoys "we'll follow up as needed" / "call the office if needed".
- MEDS_PRN_TRANSCRIPT must contain counts only. Any drug name or sig in that payload is a PHI fail.

Do not fail the PR for these (all reproduce on the base branch)
- "Acetaminophen 650 mg" resolving to "Dologen 650 mg-2 mg", a combination product.
- Lorazepam's shower dose resolving to "Ativan 1 mg" when no strength was dictated.
- The same PRN appearing as both a medication-list entry and a prescription when Nabla's own note
  lists it in both places — check from_transcript to tell that apart from a scope regression.

